PR #21, opened 2025-02-05
docs(tutorials): replace legacy tutorials with updated videos

dskuza:
- Moves the "Tutorials" section from the the "Home" sidebar into its own tab element.
- Adds a new "Learn Rive" page with links to an updated tutorial and the Rive 101 playlist
- Deletes legacy tutorials


HayesGordon (approved):
LGTM but depends on internal discussion if everyone is happy

PR #25, opened 2025-02-06
fix(runtimes): update external react-native github links

HaraldNordgren:
The links are not pointing to anything, making the experience annoying. Is it on purpose or a mistake?


dskuza (approved):
Nice find, thanks for the fix!

dskuza:
This is now live 🙌

PR #30, opened 2025-02-08
Replace all CodeSandbox examples with new forks

lancesnider:
The existing CodeSandbox links live in a few different accounts, making them hard to keep track of. 

This PR links all of the demos to the new CodeSandboxes that like in the Rive account: ttps://codesandbox.io/dashboard/sandboxes/Rive%20Docs?workspace=ws_XXze7Vt5ZRuDr9J37bb5ub

This PR also includes a bunch of auto-formatting.


HayesGordon (approved):
LGTM! Thanks for doing this

PR #32, opened 2025-02-11
docs(runtimes): add notes around ios resource usage

dskuza:


runtimes/ios-macos/resource-usage.mdx
```diff
@@ -0,0 +1,18 @@
...
+description: ''
+---
+
+When comparing resource usage between Rive and other libraries, you may encounter differences in CPU and memory usage.
+
+An important note is that Rive uses Metal APIs directly over other APIs and frameworks like Core Animation to be able to adjust its usage for best performance with Rive.
+
+<Note>Lottie is an example of a library that uses Core Animation.</Note>
+
+For libraries using Core Animation, logic and rendering is managed in a separate process, the "Render Server" (`backboardd`). In doing so, CPU and memory usage isn't reported by the app process itself, and instead is reported by `backboardd`.
```
HayesGordon, line 12: How about making this paragraph and the one below it's own section under it's own header: "Libraries using Core Animation"

HayesGordon: hmm nevermind

HayesGordon (approved):
LGTM

Maybe we can change "Rive may use more" and "Rive may use less", and instead say "To get an accurate representation of the actual memory and CPU used, consider..."

Later it will be great if we can link to actual data, or a Github repo, where people can measure themselves. Comparing us to some of these libraries.

runtimes/ios-macos/resource-usage.mdx
lancesnider: The commas confused me at first. Maybe instead:

> ...in a separate process called the "Render Server"

lancesnider (approved):

HayesGordon (approved):
Approving again after the last changes! Looks great!

PR #36, opened 2025-02-12
project: update usage of ios to apple

dskuza:
- Renames the `ios-macos` directory to `apple`
- Adds redirects for the main Apple page
- Adds wildcard redirects for all remaining Apple pages
- First pass at replacing all usage of "iOS" with "Apple platforms"


runtimes/apple/migrating-from-5.x.x-to-6.x.x.mdx
```diff
@@ -16,4 +16,4 @@ description: 'Migrating guide from &lt; 6.x'
 
 ## Package Size
 
-Rive's iOS runtime is now \~57% smaller, at \~3.3mb, compared to \~7.6mb prior to v6.0.0.
\ No newline at end of file
+Rive's Apple runtime is now \~57% smaller, at \~3.3mb, compared to \~7.6mb prior to v6.0.0.
```
HayesGordon, line 19: Maybe this is only the iOS build? So we should keep this iOS (Apple)?

runtimes/choose-a-renderer/overview.mdx
```diff
@@ -36,29 +36,29 @@ flutter run --no-enable-impeller
...
 
 <Tabs>
 
-    <Tab title="iOS/macOS">
+    <Tab title="Apple">
         
         ## Starting Version 
 
-        The Rive Renderer was introduced in the iOS runtime starting at **v5.3.2**, however, we recommend installing the latest version of the dependency to get the latest updates. See the [CHANGELOG](https://github.com/rive-app/rive-ios/blob/main/CHANGELOG.md) for details on the latest versions. Starting in v6.0.0, Rive Renderer is used as the default on iOS / macOS.
+        The Rive Renderer was introduced in Apple runtimes starting at **v5.3.2**, however, we recommend installing the latest version of the dependency to get the latest updates. See the [CHANGELOG](https://github.com/rive-app/rive-ios/blob/main/CHANGELOG.md) for details on the latest versions. Starting in v6.0.0, Rive Renderer is used as the default on Apple platforms.
```
HayesGordon, line 55: This is now already a year old. I think we can just say it was introduced in v6.0.0, but we recommend to always use the latest

runtimes/choose-a-renderer/overview.mdx
```diff
@@ -103,10 +103,10 @@ Your collaboration helps us refine and enhance the Rive Renderer to make it more
 
         Options:
 
-        - **iOS**: `Skia` (default), `Rive`, and `CoreGraphics`
+        - **Apple**: `Skia` (default), `Rive`, and `CoreGraphics`
```
HayesGordon, line 106: We should change this line, Rive is now the default

runtimes/choose-a-renderer/overview.mdx
```diff
@@ -103,10 +103,10 @@ Your collaboration helps us refine and enhance the Rive Renderer to make it more
...
         - **Android**: `Skia` (default), `Rive`, and `Canvas`
 
-        In the future, `Rive` will be the default for both iOS and Android.
+        In the future, `Rive` will be the default for both Apple platforms and Android.
```
HayesGordon, line 109: Reminder that we should make updates to the docs once Android drops skia

HayesGordon:
Looks good! I left some comments that I think needs to be addressed before merging.

Another general comment is that in some places, we say "Apple platform runtime" and in others, just "Apple runtime". When do we use one over the other?

dskuza:
@HayesGordon Updated! Re: platforms vs runtimes - maybe platforms is the wrong word, and devices is better. However, that should _only_ be if we're talking about device support, otherwise, it should be runtimes. At least, that's my initial thought. Good callout.

lancesnider (approved):

HayesGordon (approved):
LGTM!

PR #46, opened 2025-02-18
docs(react): fix nested inputs sample code

HayesGordon:


runtimes/state-machines.mdx
```diff
@@ -677,9 +677,13 @@ If the nested artboard named **FX Component** was inside the **Menu** artboard,
     To set the **Volume** input for the above example:
 
     ```tsx
-    const rive = new Rive({...});
+    const {rive, RiveComponent} = useRive({...});
```
ErikUggeldahl, line 680: Nit: Is the `RiveComponent` here necessary, or could it just be `_`?

ErikUggeldahl (approved):

runtimes/state-machines.mdx
HayesGordon: It's not needed, but you'll always want to access the RiveComponent as well when you do `useRive` and if you just access `rive`, but later do `RiveComponent` you'll have two separate instances.

So want to avoid that confusion!

PR #48, opened 2025-02-18
docs(apple): additional verbiage cleanup

HayesGordon:


HayesGordon:
@dskuza does it make sense to also have the images live under an "Apple" folder? So that we use the same wording everywhere?

ErikUggeldahl (approved):

dskuza:
> @dskuza does it make sense to also have the images live under an "Apple" folder? So that we use the same wording everywhere?

@HayesGordon Yes!

PR #50, opened 2025-02-20
docs: expand notes on rive renderer only features

HayesGordon:


dskuza (approved):
Approving the copy, but the broken links status check is failing! Good to merge once that's fixed.

PR #57, opened 2025-03-04
feat: Implement data binding overview docs

ErikUggeldahl:


editor/data-binding/overview.mdx
```diff
@@ -0,0 +1,138 @@
...
+# What is Data Binding?
+
+Data binding is a powerful way to create reactive connections between editor elements, data, and code. For instance, you might:
+- Bind the color of an icon to a color data property that can be adjusted by an engineer at runtime
+- Bind the X-position of an animated object so that it can be followed with an offset by another object
+- Listen for a click event on two buttons to increment and decrement a counter
+
+# Intro Videos
+
+The following series of videos introduce data binding as a guided walk through the editor, which will help solidify the below concepts.
```
HayesGordon, line 17:
```suggestion
The following video series introduces data binding through a guided walkthrough of the editor, which will help solidify the concepts below.
```

editor/data-binding/overview.mdx
```diff
@@ -0,0 +1,138 @@
...
+
+Data binding introduces a number of concepts that you will need to familiarize yourself with. The names of these concepts are loosely derived from the Model, View, Viewmodel (MVVM) pattern in software development.
+
+### Editor Element
+
+For the purposes of data binding, an "editor element" simply refers to an editable UI element in the editor with a value that can have a binding attached to it.
+
+### View Model
+
+A view model is a blueprint for a collection of data. Developers might think of this as similar to a class in object-oriented programming. View models typically describe all of the associated data with a given use case - commonly one per artboard. View models themselves don't have concrete values. For that, you must have an instance. See below.
```
HayesGordon, line 55: For the "See below" should we just link "instance" to the correct heading?

editor/data-binding/overview.mdx
```diff
@@ -0,0 +1,138 @@
...
+
+### View Model Instance
+
+A view model instance is the living version of a view model with actual values. Developers might think of this as similar to a class instance in object-oriented programming. Instances have the same properties as the view model they are derived from, except now each of these properties has a living value that can change over time.
+
+You may create as many instances as you'd like from a given view model. Each can be given a unique name associated with what those values represent. Each can have different initial values for its properties, representing a design-time configuration. For example, if you had a menu with three buttons with icons: 🏠 Home, 👤 Profile, and ❓ About, you might have a single artboard representing the menu item, but three view model instances, each with the menu item's label and associated icon, that can be applied to that artboard to configure the buttons.
+
+Artboards are assigned an instance in order to have populate the data bindings. Changing which instance is applied will change the initial state of the properties and all associated bound elements.
```
HayesGordon, line 67:
```suggestion
Artboards are assigned an instance to populate the data bindings. Changing which instance is applied will change the initial state of the properties and all associated bound elements.
```

editor/data-binding/overview.mdx
```diff
@@ -0,0 +1,138 @@
...
+
+### View Model
+
+A view model is a blueprint for a collection of data. Developers might think of this as similar to a class in object-oriented programming. View models typically describe all of the associated data with a given use case - commonly one per artboard. View models themselves don't have concrete values. For that, you must have an instance. See below.
+
+### View Model Property
+
+A view model property is one piece of data within a view model. Developers might think of this as similar to a field in object-oriented programming. Properties have a data type which is selected when they are created and a name which can be referenced in code. Each property can be bound to different editor elements of the same type.
+
+### View Model Instance
```
HayesGordon, line 61: We should include a bit about instancing at runtime:
- At runtime you can access an already created instance that is exported in the file
- Or you can create a brand new instance (default values) through code and bind that

editor/data-binding/overview.mdx
```diff
@@ -0,0 +1,138 @@
...
+
+Enum properties can be either a "system" enum, in which case they represent a fixed set of options in the editor, such as the "Horizontal Align" options, or a "user defined" enum, in which case they can represent any fixed set of options applicable to your use case.
+
+### Converter
+
+A converter is a general purpose way of transforming a binding's value when it is applied. These transformations might involve changing its type. For instance, the "Convert to String" converter can be used to convert a numerical binding to text, so that an object's X position could be applied to a text run.
+
+# Comparing to Existing Features
+
+Data binding fills some of the same roles as existing features in Rive. In general, it is considered a more powerful alternative to both inputs and events, and we recommend you adopt it for most use cases going forward. However, that does not mean that you need to retrofit existing files - these features will be supported for quite some time yet. Additionally, there is some overlap with constraints.
```
HayesGordon, line 101: - "these features will be supported for quite some time yet" - let's scratch this, don't want to start a panic. Currently, we don't have plans to deprecate.
- "Additionally, there is some overlap with constraints." - Same thing, don't want to give the impression we're getting rid of this


I would call out here that the following are good places to look into updating existing implementations. If you are:
- setting nested inputs at runtime
- setting nested text runs at runtime
- bubbling up nested events to get reported at runtime

editor/data-binding/overview.mdx
```diff
@@ -0,0 +1,138 @@
...
+
+### Type Support
```
HayesGordon, line 103: Should we tease the properties that will be coming in the future @bodymovin?

HayesGordon: If not, reminder to everyone that we will need to update this list

editor/data-binding/overview.mdx
```diff
@@ -0,0 +1,138 @@
...
+| Booleans               | ✅     | ✅     | ✅                     |
+| Triggers               | ✅     | ❌     | ✅                     |
+| Strings                | ❌     | ✅     | ✅                     |
+| Enumerations           | ❌     | ❌     | ✅                     |
+| Colors                 | ❌     | ❌     | ✅                     |
+| Nesting                | ❌     | ❌     | ✅                     |
+
+### State Machine Inputs
+
+Before data binding, state machine inputs were the primary way for engineers to affect designs. They formed the "input" side of the contract with design. View model properties are a more flexible system.
```
HayesGordon, line 118: Great overview

editor/data-binding/overview.mdx
```diff
@@ -0,0 +1,138 @@
...
+| Strings                | ❌     | ✅     | ✅                     |
+| Enumerations           | ❌     | ❌     | ✅                     |
+| Colors                 | ❌     | ❌     | ✅                     |
+| Nesting                | ❌     | ❌     | ✅                     |
+
+### State Machine Inputs
+
+Before data binding, state machine inputs were the primary way for engineers to affect designs. They formed the "input" side of the contract with design. View model properties are a more flexible system. 
+
+Inputs can only be used to drive state machine transitions, where data binding can be used to drive most editor elements in Rive.
```
HayesGordon, line 120:
```suggestion
Inputs can only be used to drive state machine transitions, whereas data binding can be used to drive most editor elements in Rive and state machine transitions.
```

editor/data-binding/overview.mdx
```diff
@@ -0,0 +1,138 @@
...
+
+Inputs must be used as-is where data-bound properties can be converted, either before being used by developers or before being applied to editor elements.
```
HayesGordon, line 122:
```suggestion
Inputs must be used as-is whereas data-bound properties can be converted, either before being used by developers or before being applied to editor elements.
```

editor/data-binding/overview.mdx
```diff
@@ -0,0 +1,138 @@
...
+
+View model properties also support both polling and listening APIs for developers, whereas inputs only support polling. This means that engineers can more naturally react to changes in data.
+
+View model properties can also be used in two features currently used by inputs, that being blended states (both Blend 1D and Blend Additive) as the mix parameter and as the receiver for listeners, e.g. setting a value on a mouse click or tap.
+
+### Events
```
HayesGordon, line 128: Maybe it needs a one-liner at the start to say that in the editor users can respond to data value changes similar to how events work, and at runtime, developers can listen to updates to any property

editor/data-binding/overview.mdx
```diff
@@ -0,0 +1,138 @@
...
+
+The counterpart to inputs, events were the primary way for engineers to receive "outputs" from designs. Data binding is a much richer channel for engineers to observe values from the Rive design.
```
HayesGordon, line 130: Same as the above comment, we'd need to expand a bit, as events were used not just at runtime but as a mechanism for nested artboards to communicate with their parent - which often required "bubbling up" and recreating and re-reporting multiple events n-levels deep. This is no longer needed as all artboards in the hierarchy have access to the same underlying data and can update the data and respond to changes as needed.

Maybe @bodymovin has a nice way to phrase this.

I think it would also be good to highlight what data binding won't be capable of doing, but events still are. If anything... Only thing that comes to my mind is more simple listeners, and audio events

editor/data-binding/overview.mdx
```diff
@@ -0,0 +1,138 @@
...
+
+Events can only be triggered by timelines, state machine transitions, or listeners. By comparison, data bound properties can be changed from any number of sources.
+
+Events can have fixed properties that fire the same value every time they are triggered. By comparison, view model inputs can carry new data each time they change, triggering a developer's listener with the new value.
```
HayesGordon, line 134: "Events can have fixed properties that" this part isn't true. You can update the value through keying an animation. But it becomes a problem when you need to bubble up this information from a n-level deep nested artboard and access that information at runtime

HayesGordon:
Did a first pass, this is great @ErikUggeldahl!

editor/data-binding/overview.mdx
ErikUggeldahl: Done!

editor/data-binding/overview.mdx
ErikUggeldahl: Done!

editor/data-binding/overview.mdx
ErikUggeldahl: Done!

editor/data-binding/overview.mdx
ErikUggeldahl: My preference is not to "sell futures", as there's enough to digest here without including unsupported features. But happy to include if people feel otherwise.

editor/data-binding/overview.mdx
ErikUggeldahl: Thanks!

editor/data-binding/overview.mdx
ErikUggeldahl: Added!

editor/data-binding/overview.mdx
ErikUggeldahl: I've added the gotcha about audio events. The nesting bit was covered by your earlier comment which also addresses nested inputs and text runs.

I realized I wish I could say that you can bind to an event to play the audio, but doesn't look like this is possible. Maybe we should have a property of type Asset which can hold an image, font, or audio?

editor/data-binding/overview.mdx
ErikUggeldahl: Ah, good to know! I've added clarification.

editor/data-binding/overview.mdx
```diff
@@ -0,0 +1,138 @@
...
+
+Artboards are assigned an instance to populate the data bindings. Changing which instance is applied will change the initial state of the properties and all associated bound elements.
+
+In order for an instance to be visible to developers, it must be marked as Exported. Otherwise, it is considered internal to the file. One reason you may want to keep it internal is if you only use the instance to test your design when it is configured with a given set of values, including edge cases.
+
+### Binding
+
+A binding is an association between a property and an editor element. For instance, you might have a property named "Name" bound to an text run's text value.
```
Tod-Rive, line 73: Should this be "bound to a text run's"...

ErikUggeldahl: Good catch! Fixed in #60

PR #59, opened 2025-03-05
fix(data binding): Address a number of PR comments

ErikUggeldahl:
Follows on from PR comments in #57. All should be addressed. Still welcoming general review and critique.

- Add consistency by always referring to engineers as developers
- Link to view model instances from view models
- Add information about instances at runtime
- Removed references to supporting inputs and events for some time
- Added coverage for events used internal to the Rive file, and extra clarification of feature coverage
- Fixed event property comparison to be keyable rather than fixed, and add nesting gotcha
- Added audio limitation to data binding
- Added a section about nesting use cases better served by data binding

### DIFF
--- a/editor/data-binding/overview.mdx
+++ b/editor/data-binding/overview.mdx
@@ -133,7 +133,7 @@ The counterpart to inputs, events were the primary way for developers to receive
 
 Events can only be triggered by timelines, state machine transitions, or listeners. By comparison, data bound properties can be changed from any number of sources.
 
-Events can have keyable properties with values that are passed when triggered. This is limited to being updated by animation keys and can be tricky to "bubble" when the animation exists on nested a artboard. By comparison, view model properties carry the most recent data each time they change, from any level of the hierarchy, triggering a developer's listener with the new value.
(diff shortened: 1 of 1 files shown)


editor/data-binding/overview.mdx
```diff
@@ -115,24 +117,36 @@ View model properties can represent more types of data compared to inputs and ev
...
 
 Events can only be triggered by timelines, state machine transitions, or listeners. By comparison, data bound properties can be changed from any number of sources.
 
-Events can have fixed properties that fire the same value every time they are triggered. By comparison, view model inputs can carry new data each time they change, triggering a developer's listener with the new value.
+Events can have keyable properties with values that are passed when triggered. This is limited to being updated by animation keys and can be tricky to "bubble" when the animation exists on nested a artboard. By comparison, view model properties carry the most recent data each time they change, from any level of the hierarchy, triggering a developer's listener with the new value.
```
HayesGordon, line 136:
```suggestion
Events can have keyable properties with values that are passed when triggered. This is limited to being updated by animation keys and can be tricky to "bubble" when the animation exists on a nested artboard. By comparison, view model properties carry the most recent data each time they change, from any level of the hierarchy, triggering a developer's listener with the new value.
```

HayesGordon (approved):
LGTM! Great work on this whole document

PR #60, opened 2025-03-06
feat: Add runtime data binding docs

ErikUggeldahl:
Continuing data binding documentation, this adds a section for runtimes.

Also adds an early access warning to the existing overview.


runtimes/data-binding.mdx
```diff
@@ -0,0 +1,169 @@
...
+Once we have a reference to a view model, it can be used to create an instance. When creating an instance, you have four options:
+
+1. Create an blank instance - Fill the properties of the created instance with default values as follows:
+
+    | Type              | Value           |
+    |-------------------|-----------------|
+    | Number            | 0               |
+    | String            | Empty string    |
+    | Boolean           | False           |
+    | Color             | #00000000       |
```
ErikUggeldahl, line 45: This might not currently be true. Need to talk with Hernan.

runtimes/data-binding.mdx
```diff
@@ -0,0 +1,169 @@
...
+        ```kotlin
+        val vm = view.controller.file?.getViewModelByName("My View Model")!!
+        val vmi = vm.createInstanceFromName("My Instance")
+
+        val numberProperty = vmi.getNumberProperty("My Number Property")
+        // Get
+        val myPropertyValue = numberProperty.value
+        // Set
+        myProperty.value = 10f
+        ```
```
ErikUggeldahl, line 135: I didn't include observing here because I haven't implemented it yet in Android 😅

runtimes/data-binding.mdx
```diff
@@ -0,0 +1,169 @@
...
+        for (i in 0 until file.viewModelCount) {
+            val vm = file.getViewModelByIndex(i)
+        }
+        ```
+    </Tab>
+</Tabs>
+
+Once we have a reference to a view model, it can be used to create an instance. When creating an instance, you have four options:
+
+1. Create an blank instance - Fill the properties of the created instance with default values as follows:
```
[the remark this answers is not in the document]
HayesGordon: The web and and the C++ implementation refers to it as just: `createInstance`

It is maybe better to be verbose and make clear what the call does. Tagging @bodymovin , as I don't think the runtimes (including C++) should diverge here

runtimes/data-binding.mdx
```diff
@@ -0,0 +1,169 @@
...
+        val vmiNamed = vm.createInstanceFromName("My Instance")
+        ```
+    </Tab>
+</Tabs>
+
+# View Model Instances
+
+Once you've used your view model to create an instance, that instance can be assigned to an artboard. This establishes the bindings set up at edit time.
+
+<Note>The initial values of the instance are not applied to their bound elements until the artboard is advanced.</Note>
```
HayesGordon, line 82: Regarding the above comment about binding the state machine as well.

We should rather say "until the state machine advances".

runtimes/data-binding.mdx
```diff
@@ -0,0 +1,169 @@
...
+        val properties = vm.properties
+        assertContains(
+            properties,
+            ViewModel.Property(ViewModel.PropertyDataType.NUMBER, "My Number Property")
+        )
+        ```
+    </Tab>
+</Tabs>
+
+View model instances have mutable properties. References to these properties can be retrieved by name. They have get, set, and observe operations. Getting or observing the value will retrieve the latest value set on that properties binding, as of the last artboard advance. Setting the value will update the value and all of its bound elements. Note that trigger properties can only be set.
```
HayesGordon, line 120: "state machine advance"

HayesGordon: You can also register callbacks on `trigger` properties (not get, but just observe when it's triggered_. You might need to double-check that Android implemented that. I also missed that behavior in my initial implementation for rive_native.

runtimes/data-binding.mdx
```diff
@@ -0,0 +1,169 @@
...
+
+<Note>After setting a property's value, the changes will not apply to their bound elements until the artboard is advanced.</Note>
```
HayesGordon, line 122: "state machine advanced" - maybe @bodymovin has thoughts about the language here. But databinding isn't usable outside of state machines (if I'm not mistaken).

runtimes/data-binding.mdx
HayesGordon: We should have a dedicated section/header for "Observability", it's a cool feature and we should highlight it

runtimes/data-binding.mdx
```diff
@@ -0,0 +1,169 @@
...
+        val myPropertyValue = numberProperty.value
+        // Set
+        myProperty.value = 10f
+        ```
+    </Tab>
+</Tabs>
+
+# Enums
+
+Enums properties come in two flavors: system and user-defined. In practice, you will not need to worry about the distinction, but just be aware that system enums are always available in all Rive files, representing options from the editor's dropdowns, where user-defined enums are those defined by a designer in the editor.
```
HayesGordon, line 141: @bodymovin to clarify, but as I understood it the system enums won't always be on every file, but it be populated in a file if that file uses system enums

runtimes/data-binding.mdx
```diff
@@ -0,0 +1,169 @@
...
+
+Enums are string typed. The Rive file contains a list of enums. Each enum in turn has a name and a list of strings.
```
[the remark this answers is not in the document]
HayesGordon: At the core level, yes, but some do host. I don't think we need to clarify, and this language is fine as  "The Rive file" could also be interpreted as the .riv

And if you did step through the hoisting, the data is accessed from the File object.

runtimes/data-binding.mdx
```diff
@@ -0,0 +1,169 @@
...
+        ```kotlin
+        val enums = view.controller.file?.enums!!
+
+        val firstEnumName = enums[0].name
+        val firstEnumFirstValue = enums[0].values[0]
+        ```
+    </Tab>
+</Tabs>
+
+# Nested Property Paths
```
HayesGordon, line 156: We should expand on the other way to access nested properties:
```dart
var eyeInstance = `viewModel.instance("person).instance("eye");
eyeInstance.color("color").value = Colors.blue;
```

Then lead into the path based API where you can access the property directly.

runtimes/data-binding.mdx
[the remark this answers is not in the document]
ErikUggeldahl: Thanks for clarifying!

runtimes/data-binding.mdx
ErikUggeldahl: Hmm, if you think so. I thought of it as a slightly better alternative to polling, but in the same vein. I can't really add the section yet, as I don't know how to talk about it without implementing it. But good to keep as a to-do.

runtimes/data-binding.mdx
```diff
@@ -0,0 +1,169 @@
...
+            val vmiIndexed = vm.createInstanceFromIndex(i)
+        }
+
+        // Create by name
+        val vmiNamed = vm.createInstanceFromName("My Instance")
+        ```
+    </Tab>
+</Tabs>
+
+# View Model Instances
```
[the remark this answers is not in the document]
ErikUggeldahl: I was not aware of this at all! In the editor, I see that the artboard has a "Data Bind" section where the user can assign an instance. I don't see an equivalent for the state machine. Though I do see the `StateMachineInstance::setDataContextFromInstance`.

What does that mean functionally?
- Does it just exist to attach the bindings for state machine transitions?
- Is this handled automatically in the editor, mirroring the artboard?
- Do we expect runtime users to bind both always?
- Does it make sense for them to bind different instances to the artboard and state machine?
  - If not, should we just make it one operation on the artboard that internally also sets it on the state machine?
    - And we could further simplify by having the C++ runtime do this automatically?

runtimes/data-binding.mdx
ErikUggeldahl: I've changed to "until the state machine advances", but I'll await further clarification on the above.

runtimes/data-binding.mdx
ErikUggeldahl: Got it. I guess I never realized because in Android, state machine advances are tied to artboard advances, so my unit tests never revealed that level of specificity.

runtimes/data-binding.mdx
ErikUggeldahl: Good call. I've also incorporated that into my unit tests.

runtimes/data-binding.mdx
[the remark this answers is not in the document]
ErikUggeldahl: The DevRels just discussed this and I think it makes sense for C++ to choose to expose either binding on the artboard or the state machine, and then the C++ runtime handles binding the other.

@bodymovin I think it was mentioned that you discussed with @Tod-Rive that these are separated for a hypothetical use case where it would be useful, but we never found a concrete reason that it would be useful for. Might it make sense to revisit this decision?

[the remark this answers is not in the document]
bodymovin: I don't think we haven't found a valid use case for having two separate instances bound to the artboard and the state machine.
But artboards don't have a reference to their state machines, so it's not possible to do a single API that takes care of both.
What we can do is let the user bind to any, but if they only bind to the state machine, internally we bind to the artboard as well.
And perhaps add a note clarifying that binding to the state machine overwrites any previous binding to the artboard.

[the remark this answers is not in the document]
ErikUggeldahl: @bodymovin Just getting to this now, thanks for the explanation. I believe it would be simplest to only expose the one API to state machines and assume it's always connected to its artboard. If we agree to this, which I believe all DevRels are onboard, what do you think about doing this at the C++ level rather than in each runtime? Would that be feasible?

Also, if we were to do this, would that mean that in order to use data binding you would need a state machine? i.e. a file with just linear animations and data binding to the artboard wouldn't work because there would be no state machine to attach it to?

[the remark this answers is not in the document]
bodymovin: yes, we can do it at the c++ level.
What I'm not fully on board is not allowing to data bind to the standalone artboard.
There are cases where an artboard might not have a state machine but its nested artboards do. Disabling data binding for those scenarios seems more limiting that what is really possible.
Is there any other reason not to expose the API on both besides simplicity / clarity for users?

ErikUggeldahl:
Added my latest pass. Added a section on auto-binding and added web examples to the best of my knowledge. I think they should be double checked.

Couple outstanding items:
- [ ] Should we breakout observability into its own section? Does it warrant it?
- [ ] Should we use different nomenclature for blank instances? I can rename my methods in Kotlin to just `instance` or similar, but how do we want to refer to them in text? The term "default" is already taken for the one marked default in the editor.

snippets/early-access.mdx
```diff
@@ -0,0 +1,5 @@
+export const EarlyAccess = ({ featureName }) => {
```
HayesGordon, line 1: Nice!

editor/data-binding/overview.mdx
```diff
@@ -119,6 +122,8 @@ View model properties can represent more types of data compared to inputs and ev
 | Colors                 | ❌     | ❌     | ✅                     |
 | Nesting                | ❌     | ❌     | ✅                     |
 
+We intend to support lists and images in following releases.
```
HayesGordon, line 125: We could consider putting these in the table and saying: "Coming soon". But not a blocker and up for discussion

runtimes/data-binding.mdx
```diff
@@ -0,0 +1,316 @@
...
+description: 'Connect your code to bound editor elements using View Models'
+---
+
+import { EarlyAccess } from '/snippets/early-access.mdx'
+
+<EarlyAccess featureName="Data Binding"/>
+
+# Overview
+
+Before engaging with the runtime data binding APIs, it is important to familiarize yourself with the core concepts presented in the [Overview](/editor/data-binding/overview).
```
HayesGordon, line 12:
```suggestion
Before engaging with the runtime data binding APIs, it is important to familiarize yourself with the core concepts presented in the [Editor Overview](/editor/data-binding/overview).
```

HayesGordon: If we wanted to be fancy, we can also have a bigger callout to the editor section by doing the following:

```
<CardGroup cols={1}>
  <Card title="Data Binding Concepts" href="./web"  icon={<svg xmlns="http://www.w3.org/2000/svg" height="100%" fill="none" viewBox="0 0 16 16" class="size-4 text-gray-500/80 dark:text-gray-400" aria-hidden="true"><path fill="currentColor" d="M7.31 7.111 2.406 5.15l4.61-1.844.328-.126a2.3 2.3 0 0 1 1.647 0l.33.126L13.93 5.15 9.024 7.112c-.55.22-1.163.22-1.712 0"></path><path stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" d="m2.405 10.911 4.906 1.963c.55.22 1.162.22 1.712 0l4.906-1.963M2.405 8.031 7.31 9.992c.55.22 1.162.22 1.712 0l4.906-1.963M2.405 5.15 7.31 7.111c.55.22 1.162.22 1.712 0l4.906-1.962-4.61-1.844-.329-.126a2.3 2.3 0 0 0-1.647 0l-.329.126z"></path></svg>}>
    An overview of core Data Binding concepts.
  </Card>
</CardGroup>
```

runtimes/data-binding.mdx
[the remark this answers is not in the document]
HayesGordon: We should decide what to do here.

runtimes/data-binding.mdx
```diff
@@ -0,0 +1,316 @@
...
+
+        // Get reference to the default view model
+        val defaultVM = file.defaultViewModelForArtboard(view.controller.activeArtboard!!)
+        ```
+    </Tab>
+</Tabs>
+
+Once we have a reference to a view model, it can be used to create an instance. When creating an instance, you have four options:
+
+1. Create an blank instance - Fill the properties of the created instance with default values as follows:
```
HayesGordon, line 62:
```suggestion
1. Create a blank instance - Fill the properties of the created instance with default values as follows:
```

runtimes/data-binding.mdx
```diff
@@ -0,0 +1,316 @@
...
+
+<EarlyAccess featureName="Data Binding"/>
+
+# Overview
+
+Before engaging with the runtime data binding APIs, it is important to familiarize yourself with the core concepts presented in the [Overview](/editor/data-binding/overview).
+
+# View Models
+
+View models describe a set of properties, but cannot themselves be used to get or set values - that is the role of view model instances.
```
HayesGordon, line 16: Could link to `view model instances` below

runtimes/data-binding.mdx
```diff
@@ -0,0 +1,316 @@
...
+            val indexedVM = file.getViewModelByIndex(i)
+        }
+
+        // Get reference to the default view model
+        val defaultVM = file.defaultViewModelForArtboard(view.controller.activeArtboard!!)
+        ```
+    </Tab>
+</Tabs>
+
+Once we have a reference to a view model, it can be used to create an instance. When creating an instance, you have four options:
```
HayesGordon, line 60: I wonder if this isn't better to be part of the View Model Instance section?

runtimes/data-binding.mdx
```diff
@@ -0,0 +1,316 @@
...
+
+        // Alternatively, apply the instance to the artboard
+        view.controller.activeArtboard?.viewModelInstance = vmi
+        ```
+    </Tab>
+</Tabs>
+
+### Auto-Binding
+
+Alternatively, you may prefer to use auto-binding. This will automatically bind the default view model of the artboard using the default instance to both the state machine and the artboard. This is useful when your design only has one configuration.
```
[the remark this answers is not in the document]
HayesGordon: Suggesting we remove that last line, as although it's true, it's not the only reason you may want to autobind. To avoid giving the impression that people should only use it if they have a single configuration.

Might also be worth calling out again that the default is determined by what was set as the default in the editor

runtimes/data-binding.mdx
```diff
@@ -0,0 +1,316 @@
...
+        }
+        // Or collect in Compose
+        val numberValue by numberProperty.collectAsState()
+        ```
+    </Tab>
+</Tabs>
+
+# Enums
+
+Enums properties come in two flavors: system and user-defined. In practice, you will not need to worry about the distinction, but just be aware that system enums are available in any Rive file that binds to a editor defined enum set, representing options from the editor's dropdowns, where user-defined enums are those defined by a designer in the editor.
```
HayesGordon, line 260:
```suggestion
Enums properties come in two flavors: system and user-defined. In practice, you will not need to worry about the distinction, but just be aware that system enums are available in any Rive file that binds to an editor-defined enum set, representing options from the editor's dropdowns, where user-defined enums are those defined by a designer in the editor.
```

runtimes/data-binding.mdx
```diff
@@ -0,0 +1,316 @@
...
+        ```kotlin
+        view.setRiveResource(
+            R.raw.my_rive_file,
+            autoBind = true,
+        )
+        ```
+    </Tab>
+</Tabs>
+
+# Properties
```
HayesGordon, line 184: Should we have a table here that is a reference to the available properties?
Maybe this can be a shared element that we use here and in the Editor Overview?

HayesGordon:
> Added my latest pass. Added a section on auto-binding and added web examples to the best of my knowledge. I think they should be double checked.
> 
> Couple outstanding items:
> 
> * [ ]  Should we breakout observability into its own section? Does it warrant it?
> * [ ]  Should we use different nomenclature for blank instances? I can rename my methods in Kotlin to just `instance` or similar, but how do we want to refer to them in text? The term "default" is already taken for the one marked default in the editor.

Thoughts:
- I think it does potentially warrant its own section for a few reasons (but should not block this first release of the docs):
    - Clear visibility that this functionality is available so that someone can find the API at a glance of the docs. But this can also be solved by having a subsection under **Properties** so that you can see observability listed there
    In the future, we may want to call out caveats regarding observability (shared VMIs, same observation across multiple files, reassigning).
    - List observability might also come with additional consideration/usage. But maybe in that case **Lists** should be it's own section with a subsection on observability (same suggestion as above)
- I like your use of blank instances in the docs, even though the API right now does not indicate that everywhere, which we should settle on. But from a documentation perspective, I like it. Saying "default" might give the impression that it is the designer-defined default values, not the blank defaults.

HayesGordon (approved):
LGTM!

runtimes/data-binding.mdx
ErikUggeldahl: I agree that a refresher table is not a bad idea. Though I do think it's unfortunately not a great one to share because of the additional "inputs and events" columns in the overview section. There might be some MDX magic that could drop those columns with a parameter, but I doubt it's worth the effort. Doing it manually for now.

editor/data-binding/overview.mdx
```diff
@@ -8,7 +8,9 @@ import { YouTube } from '/snippets/youtube.mdx'
 import { EarlyAccess } from '/snippets/early-access-feature.mdx'
 
 
-<EarlyAccess featureName="Data Binding"/>
+<EarlyAccess featureName="Data Binding">
+  As data binding continues to develop additional features, we suggest you continue to refer to this page for changes.
```
HayesGordon, line 12: 🙌

PR #66, opened 2025-03-07
fix: Data binding title

ErikUggeldahl:
As requested on Slack - our search results just show "Overview". This will better capture the article's contents.


dskuza (approved):
Nice. Good intermediate solution. Next step for me is to see if / how we can make this change without affecting the sidebar, or if we can affect SEO without having to change our title.

ErikUggeldahl:
Closing in favor of #67

Issue #72, opened 2025-03-18
Data Binding Early Access notice


KnightOwl-Git:
Path: /editor/data-binding/overview

It would be nice to have a notice at the top of the page that Data Binding is still in early access, to prevent confusion with newer users.

dskuza:
This has been merged in #74.

PR #74, opened 2025-03-19
docs(data-binding): add early access callout

dskuza:
Adds a reusable early access feature snippet, first used in the Data Binding overview.

Addresses #72.


HayesGordon (approved):
LGTM

PR #96, opened 2025-04-09
docs(account-admin): add custom bucket documentation 

umberto-sonnino:
First stab at adding docs for custom bucket configurations

### DIFF
--- a/account-admin/bring-your-own-bucket.mdx
+++ b/account-admin/bring-your-own-bucket.mdx
@@ -1,0 +1,117 @@
+---
+title: "Bring Your Own S3 Bucket"
+description: "Configure your own AWS S3 bucket to work with Rive"
+---
+
+**Note:** This feature is available exclusively to Enterprise plan customers. If you're interested in upgrading to unlock this capability, [contact us](https://rive.app/enterprise).
+
+## What is a Custom S3 Bucket?
+
+A custom S3 bucket allows Enterprise customers to store their Rive file and asset data in their own AWS environment instead of Rive's default storage. This premium feature gives you greater control over your data, allowing you to:
+
+- Apply your organization's security and compliance policies
+- Keep data within your existing AWS infrastructure
+- Integrate with your existing backup and disaster recovery processes
+- Monitor and audit all access to your bucket using CloudTrail
+
+If you have existing data in Rive that you would like migrated to your custom S3 bucket after upgrading to Enterprise, our team can help coordinate the migration process.
+
+## Configuration
+
+<Steps>
+  <Step title="Create an S3 bucket in your AWS Account">
+    Follow the [AWS documentation](https://docs.aws.amazon.com/AmazonS3/latest/userguide/create-bucket-overview.html) to create a new S3 bucket:
+    
+    - Choose a unique name for your bucket <br />
+      **Note:** Use that name in place of `BUCKET_NAME` for the rest of this document
+    - Configure basic bucket settings
+      - Block public access
+      - You can choose to leave everything as default or you can decide to enable or customize:
+        - versioning (disabled by default)
+        - encryption - by default buckets and new objects are encrypted by Amazon's S3 managed keys (SSE-S3) which uses AES256
+        - tags
+  </Step>
+
+  <Step title="Create an IAM Policy">
+    - In the AWS Console, under IAM / Policies click on _"Create policy"_
+    - Select the _"JSON"_ Policy Editor view and paste the following:
+
+    ```json
+    {
+      "Statement": [
+        {
+          "Action": [
+            "s3:GetObject",
+            "s3:PutObject",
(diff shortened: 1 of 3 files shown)


account-admin/byob.mdx
```diff
@@ -0,0 +1,85 @@
+---
+title: "Bring Your Own S3 Bucket"
+description: ""
+---
+
+Users of the Enterprise plan can create their own S3 buckets and configure them to work with Rive. <br />
+[Contact us](https://rive.app/enterprise) to sign up for an Enterprise plan.
+
```
avivian, line 8: I think we should also mention that customers will need to get in touch with Rive if they want to set up a custom bucket, as there's setup we'll need to do on our side. We sort of imply that towards the end with "Pass the following to your Rive representative", but we should make it explicit from the start.

We should also mention that if you have existing data in Rive you would like to have migrated to your S3 bucket, they should get in contact with us regarding that migration

avivian: Also, we don't explain anywhere here "why" a customer would want a "custom bucket", or what it even is. I'm not sure if that is a separate doc or not

dskuza (approved):
Looks good! Reads super well to someone who doesn't really understand S3 (i.e me 😅). I think @avivian should give a ✅ before this is fully merged.

PR title nitpick: it should follow conventional commit styling similar to `docs(account-admin): add page for custom bucket` (since that'll by default become the squash n' merge commit message).

account-admin/byob.mdx
```diff
@@ -0,0 +1,119 @@
...
+---
+
+## What is a Custom S3 Bucket?
+
+A custom S3 bucket allows Enterprise customers to store their Rive assets in their own AWS environment instead of Rive's default storage. This premium feature gives you greater control over your data, allowing you to:
+
+- Apply your organization's security and compliance policies
+- Keep data within your existing AWS infrastructure
+- Maintain direct ownership of storage costs and capacity
+- Implement custom lifecycle policies for data retention
```
avivian, line 13: hmm, we only want this on temporary data... Probably not a selling point

account-admin/byob.mdx
```diff
@@ -0,0 +1,119 @@
...
+description: "Configure your own AWS S3 bucket to work with Rive"
+---
+
+## What is a Custom S3 Bucket?
+
+A custom S3 bucket allows Enterprise customers to store their Rive assets in their own AWS environment instead of Rive's default storage. This premium feature gives you greater control over your data, allowing you to:
+
+- Apply your organization's security and compliance policies
+- Keep data within your existing AWS infrastructure
+- Maintain direct ownership of storage costs and capacity
```
avivian, line 12: I wouldn't say either of these are selling points

account-admin/byob.mdx
```diff
@@ -0,0 +1,119 @@
...
+<Steps>
+  <Step title="Create an S3 bucket in your AWS Account">
+    Follow the [AWS documentation](https://docs.aws.amazon.com/AmazonS3/latest/userguide/create-bucket-overview.html) to create a new S3 bucket:
+    
+    - Choose a unique name for your bucket <br />
+      **Note:** Use that name in place of `BUCKET_NAME` for the rest of this document
+    - Configure basic bucket settings
+      - Block public access
+      - You can choose to leave everything as default or you can decide to enable or customize:
+        - versioning (disabled by default)
```
avivian, line 32: I wouldn't get them to turn this on... It will massively increase costs based on how we write over revisions all the time

account-admin/byob.mdx
```diff
@@ -0,0 +1,119 @@
+---
+title: "Bring Your Own S3 Bucket"
+description: "Configure your own AWS S3 bucket to work with Rive"
+---
+
+## What is a Custom S3 Bucket?
+
+A custom S3 bucket allows Enterprise customers to store their Rive assets in their own AWS environment instead of Rive's default storage. This premium feature gives you greater control over your data, allowing you to:
```
avivian, line 8:
```suggestion
A custom S3 bucket allows Enterprise customers to store their Rive file and asset data in their own AWS environment instead of Rive's default storage. This premium feature gives you greater control over your data, allowing you to:
```

umberto-sonnino: Yeah I was torn by this too, but then I thought that a File is an asset in some way..? But, since you noticed this too, I'll change it up

account-admin/byob.mdx
```diff
@@ -0,0 +1,117 @@
+---
+title: "Bring Your Own S3 Bucket"
```
HayesGordon, line 2: @umberto-sonnino @avivian nitpick but should we name the file `bring-your-own-bucket.mdx`, as that is what will be the URL address. For SEO purposes.

umberto-sonnino: Sure thing, I'll update it

PR #98, opened 2025-04-09
docs(data-binding): add ios api example usage

dskuza:


runtimes/data-binding.mdx
```diff
@@ -64,6 +64,24 @@ To begin, we need to get a reference to a particular view model. This can be don
...
+        // Data binding view model by index
+        for index in 0..<file.viewModelCount {
+            let viewModelByIndex = file.viewModel(at: index)
+        }
+
+        // Default data binding view model for an artboard
+        guard let artboard = riveViewModel.riveModel?.artboard else { return }
+        let viewModelForArtboard = file.viewModel(for: artboard)
+        ```
+    </Tab>
```
ErikUggeldahl, line 84: For consistency with the other pages, I think it would make sense to put the Apple tab before Android.

runtimes/data-binding.mdx
```diff
@@ -64,6 +64,24 @@ To begin, we need to get a reference to a particular view model. This can be don
         val defaultVM = file.defaultViewModelForArtboard(view.controller.activeArtboard!!)
         ```
     </Tab>
+    <Tab title="Apple">
+        ```swift
+        let riveViewModel = RiveViewModel(...)
+        guard let file = riveViewModel.riveModel?.riveFile else { return }
```
ErikUggeldahl, line 70: For the sake of simplicity in these examples would it makes sense (here and elsewhere) to use force unwrapping with `!`? That's how I handled mine - leaving error handling approaches to the user.

runtimes/data-binding.mdx
```diff
@@ -194,6 +252,18 @@ Alternatively, you may prefer to use auto-binding. This will automatically bind
         )
         ```
     </Tab>
+    <Tab title="Apple">
+        ```swift
+        let riveViewModel = RiveViewModel(...)
+        riveViewModel.riveModel?.enableAutoBind { instance in
+            // Store a reference to `instance` to later access properties
+            // The instance may change as state machines and artboards change
+        }
```
ErikUggeldahl, line 261: Interesting. This is different from how I've done it. Because of instance caching, if the user wants to get the instance that was auto bound, it would the default instance for the active artboard. Is that not the case for you?

runtimes/data-binding.mdx
```diff
@@ -290,6 +370,29 @@ View model instances have mutable properties. References to these properties can
...
     </Tab>
+    <Tab title="Apple">
+        ```swift
+        let riveViewModel = RiveViewModel(...)
+        guard let instance = riveViewModel.riveModel?.riveFile.viewModelNamed("...")?.createDefaultInstance() else { return }
+        
+        // An example of getting a string property
+        let stringProperty = instance.stringProperty(fromPath: "...")
+        // Updating its value
+        stringProperty?.value = "Hello, Rive"
```
ErikUggeldahl, line 381: Might make sense to also include a simple get operation using `value` without subscribing.

runtimes/data-binding.mdx
```diff
@@ -356,4 +459,17 @@ View models can have properties of type view model, allowing for arbitrary nesti
...
     </Tab>
+    <Tab title="Apple">
+        ```swift
+        let riveViewModel = RiveViewModel(...)
+        guard let instance = riveViewModel.riveModel?.riveFile.viewModelNamed("...")?.createDefaultInstance() else { return }
+
+        let nestedNumberByChain = instance
+                                    .viewModelInstanceProperty(fromPath: "Nested View Model")
+                                    .viewModelInstanceProperty(fromPath: "Another Nested View Model")
+                                    .numberProperty(fromPath: "Nested View Model/Another Nested View Model")
```
ErikUggeldahl, line 470: Is this correct? Shouldn't the number property have a name? Seems to be repeating the path from above.

runtimes/data-binding.mdx
```diff
@@ -356,4 +459,17 @@ View models can have properties of type view model, allowing for arbitrary nesti
...
+
+        let nestedNumberByPath = instance.numberProperty(fromPath: "Nested View Model/Another Nested View Model")
```
ErikUggeldahl, line 472: Same here, should this terminate in a name?

ErikUggeldahl (changes requested):
Nice! Just a few things for you to consider.

runtimes/data-binding.mdx
dskuza: The iOS runtime doesn't (currently) cache instances, but since you can set a new artboard at runtime, there's a possibility that you'll get a different instance rebound when the artboard is changed to something different _because_ the new artboard has a different default instance. Perhaps autoBind works a little differently to me where when you update the artboard _or_ state machine, you _may_ get a new instance.

dskuza:
@ErikUggeldahl I believe I addressed all of your comments!

HayesGordon (approved):
LGTM

runtimes/data-binding.mdx
ErikUggeldahl: I see. So this may change if you start caching, but good to know.

runtimes/data-binding.mdx
```diff
@@ -357,3 +503,48 @@ View models can have properties of type view model, allowing for arbitrary nesti
         ```
     </Tab>
 </Tabs>
+
+# Listeners
+
+Listeners are a way of adding observability to your data binding properties. Using listeners, you will be notified when your property changes are applied by a state machine, whether a new value has been explicitly set or if the state machine itself has caused a change. Adding a listener to trigger properties is a new way to notify you that some event has occurred.
```
ErikUggeldahl, line 509:
```suggestion
# Observability

You can observe changes over time to property values, either by using listeners or a platform equivalent method. Once observed, you will be notified when the property changes are applied by a state machine advance, whether that is a new value that has been explicitly set or if the value was updated as a result of a binding. Observing trigger properties is an alternative method to receive events from the editor, as compared to [Rive Events](/runtimes/rive-events).
```

ErikUggeldahl: Generalizing the language here. Also, would this section make more sense as a sub-section of properties, immediately after showing how to get values, since this is just a more advanced use case of that?

ErikUggeldahl (approved):
Approved but please consider the suggestion before merging. Android doesn't use the language of listeners anywhere, so I've generalized it a bit.

PR #104, opened 2025-04-16
docs(react): add note about data binding support

HayesGordon:


runtimes/data-binding.mdx
```diff
@@ -9,6 +9,14 @@ import { EarlyAccess } from '/snippets/early-access-feature.mdx'
   As data binding continues to develop additional features, we suggest you continue to refer to this page for changes.
 </EarlyAccess>
 
+export const DataBindingReact = () => {
+  return (
+    <Info>
+    Hooks for data binding will soon be available in the Rive React package. In the meantime, you can use the Web API in React. Check back here for updates.
```
dskuza, line 15: > you can use the Web API in React.

Should we link to some other documentation here? Not sure _where_ since this snippet is within the context of the data binding page, though.

dskuza (approved):
Just a quick comment, but nice snippet!

runtimes/data-binding.mdx
HayesGordon: These are used in the individual Tabs all over the place, so I think people should realise to hop on over to the Web tab for more info

PR #117, opened 2025-04-17
docs(unity): initial data binding guide

damzobridge:
This PR adds some initial documentation for DB in Unity, with a dedicated page in the Unity section and tabs in the main runtime DB docs.


game-runtimes/unity/data-binding.mdx
```diff
@@ -0,0 +1,158 @@
+---
+title: 'Data Binding'
+description: 'Binding Rive Data to Unity and vice versa'
+---
+
+# Quick Start
+
+Data binding allows you to establish connections between your Unity code and your Rive graphics. This creates a two-way reactive system where changes in your code affect the Rive design, and changes in the design can trigger responses in your code.
```
dskuza, line 8:
```suggestion
Data Binding allows you to establish connections between your Unity code and your Rive graphics. This creates a two-way reactive system where changes in your code affect the Rive design, and changes in the design can trigger responses in your code.
```

I think we should ask (Guido) if we should be referencing this as the feature name itself (which is Data Binding) vs other casing (Data binding, data binding, etc). This would affect naming across this doc + other pages.

dskuza (approved):
Just one comment for docs moving forward, but not really a blocker to get this out the door.

PR #126, opened 2025-04-30
docs(unity): update examples of databinding observability

damzobridge:
This PR:
- Updates the docs to show the new DB property observability callbacks, which now include the new value. 
- Updates the examples to use the correct property for getting an artboard's default ViewModel.
- Adds an example of autobinding with the low-level API


Tod-Rive:
Looks good. I wonder if I should add Unreal examples alongside the others in the non-game section.

Tod-Rive (approved):

damzobridge:
> Looks good. I wonder if I should add Unreal examples alongside the others in the non-game section.

Thanks @Tod-Rive ! I found myself repeating a lot of the stuff mentioned on the main databinding page, so I decided to do a general Unity workflow overview in the Unity section instead and then just link to the main page for the data binding concepts.

PR #141, opened 2025-05-15
Convert all links to be root relative

lancesnider:
Root relative links open faster and work even if you move the page: https://mintlify.com/docs/text#linking-to-pages

I tested each of them as  I edited.


dskuza (approved):
This is great! Thanks :D We should maybe add to the CONTRIBUTING file that links should be root-relative.

lancesnider:
> This is great! Thanks :D We should maybe add to the CONTRIBUTING file that links should be root-relative.

Good call. Done.

PR #142, opened 2025-05-16
docs(web): updated RiveParameters `file` to `riveFile`

stormmuller:
https://github.com/rive-app/rive-wasm/blob/master/js/src/rive.ts#L1351

image


dskuza (approved):
Nice catch, thanks!

PR #146, opened 2025-05-21
docs: update and add mcp to editor section

dskuza:


docs.json
```diff
@@ -212,6 +212,12 @@
                   "editor/share-links/framer-and-rive"
                 ]
               },
+              {
+                "group": "MCP",
```
lancesnider, line 216: Does everyone know what this is? I didn't before rive.  Do we need to say mcp (ai) or something?

editor/mcp/integration.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+import { EarlyAccess } from '/snippets/early-access-feature.mdx'
+
+<EarlyAccess featureName="MCP">
+   This is initially available in the Early Access version of the Mac Desktop app. Windows support is coming.
+</EarlyAccess>
+
+## Getting started
+
+You can connect the Rive Editor to AI tools through MCP (Model Context Protocol). The first set of tools are designed let AI handle repetitive tasks, like creating complex View Models, State Machines with hundreds of states/layers, Layouts, Shapes, and more.
```
lancesnider, line 14: Designed "to" let...

editor/mcp/integration.mdx
```diff
@@ -0,0 +1,103 @@
...
+        }
+    }
+  }
+  ```
+  </Step>
+  <Step title="Configure MCP">
+    Open Terminal and type the following commands:
+    ```bash
+    # In a standard Cursor installation
+    cp /path/to/mcp.json ~/.cursor
```
lancesnider, line 47: Terminal is scary to non devs. Can they just create and drop it into the relevant folder?

lancesnider (approved):

editor/mcp/integration.mdx
dskuza: It's tricky - on macOS, `.cursor` will be a hidden directory and not immediately visible in Finder unless you also explicitly enable showing hidden files and folders.

docs.json
dskuza: My gut is telling me this is geared towards people who _know_ what MCP is, since it's a _bit_ of an involved process right now, and not just a site you'll visit for AI interactivity. But, if users are finding this confusing, we can always change it; I don't think updating the group would require a redirect.

dskuza: @RobertHaynie Has there been discussion on MCP vs something like MCP (AI)?

editor/mcp/integration.mdx
dskuza: @lancesnider I updated the content here a _little_ bit… can you 👀 again?

PR #147, opened 2025-05-22
docs: split early access and early access feature into separate snippets

dskuza:
- Adds usage of new EarlyAccess snippet to MCP Integration


HayesGordon (approved):
LGTM

PR #155, opened 2025-05-29
Add more optimization tips to the Best Practices page

lancesnider:

PR #158, opened 2025-06-04
docs: additional best practice updates

lancesnider:
I merged this before Gordon had a chance to review: https://github.com/rive-app/rive-docs/pull/155/files. This addresses his comments:

### DIFF
--- a/getting-started/best-practices.mdx
+++ b/getting-started/best-practices.mdx
@@ -51,7 +51,7 @@ Working directly in Rive is generally preferable as it allows for file optimizat
 
 ### Layer Blend Modes for Web
 
-Blend modes are particularly expensive on the web due to limitations in how WebGL handles framebuffer access. To apply a blend mode, Rive must copy rendered pixels into a separate texture before compositing, which introduces significant performance and memory overhead.
+Blend modes are particularly expensive on the web because WebGL doesn’t expose mechanisms for accessing the framebuffer. To apply a blend mode, Rive must copy rendered pixels into a separate texture before compositing, which introduces significant performance and memory overhead.
 
 While there are efforts underway to improve this through new WebGL capabilities, these solutions are still pending broader support. Until then, it’s best to use blend modes sparingly in web projects to ensure optimal performance.
 


getting-started/best-practices.mdx
```diff
@@ -55,7 +55,7 @@ Working directly in Rive is generally preferable as it allows for file optimizat
 
 Clipping artboards is generally fine, but if you're experiencing performance issues, it's worth minimizing their use. Clipping can be computationally expensive, as the renderer must evaluate every object—including nested artboards—to determine pixel visibility. Instead, consider applying clips to specific objects or groups within the artboard.
 
-You can also safely remove the clip from the main artboard itself, since nothing outside the Rive instance will be rendered at runtime.
```
lancesnider, line 58: > Not true for all the runtimes! But, for now it's more common for this to be the case

getting-started/best-practices.mdx
```diff
@@ -55,11 +55,11 @@ Working directly in Rive is generally preferable as it allows for file optimizat
...
+In most cases you can safely remove the clip from the main artboard itself, since nothing outside the Rive instance will be rendered at runtime.
 
 #### Unused Artboards
 
-Unused artboards are still included in the compiled .riv file. While they don't significantly impact file size, it's best practice to remove any artboards that aren't used in your animation to keep your project clean and efficient.
```
lancesnider, line 62: > It will also still be parsed and loaded at runtime. Which means the first time you decode the Rive File it'll be more expensive.

getting-started/best-practices.mdx
```diff
@@ -92,7 +92,7 @@ See our documentation on [Loading Assets](/runtimes/loading-assets). This featur
 
 ### Caching your .riv
 
-If you’re going to use the same Rive file in multiple places throughout your page or app, it’s possible to [cache the .riv](/runtimes/caching-a-rive-file) so that your users only download it once.
```
lancesnider, line 95: > i would change this to not just say "download it", because downloading an asset is only relevant for the web runtime really.

> The real benefit is that you don't have to decode and parse the .riv file multiple times. Spawning an artboard instance from an already decoded .riv file is much faster than first decoding the .riv file and then creating an artboard instance.

getting-started/best-practices.mdx
```diff
@@ -49,17 +49,23 @@ While Rive provides a Lottie converter for easy .riv file export, recreating gra
 
 Working directly in Rive is generally preferable as it allows for file optimization based on your animation requirements. For instance, using bones and constraints to create rigs will result in fewer keys in the animation compared to converting directly from Lottie.
 
+### Layer Blend Modes for Web
```
HayesGordon, line 52: This is mostly true for WebGL2 (Rive Renderer) I believe!

I've been meaning to add a note about this in the web section of our docs too. Basically if you choose to use the Rive Renderer to speed up other areas, you need to account for this blend mode perf issue.

This is something that is not a problem if you are using the experimental chrome flags to enable pixel local storage. It's only a problem with the MSAA fallback and not something we can fix afik.

PR #165, opened 2025-06-13
docs(db): mark image and list properties as early access

dskuza:


runtimes/data-binding.mdx
```diff
@@ -1394,7 +1394,9 @@ You can observe changes over time to property values, either by using listeners
 
 ### Images
 
-Image properties let you set and replace raster images at runtime, with each instance of the image managed independently. For example, you could build an avatar creator and dynamically update features — like swapping out a hat — by setting a view model’s image property.
+<EarlyAccessFeature featureName="Image property support"/>
```
dskuza, line 1397: I decided to use "Image property support" since the following word is "is", unless we want "Image properties is in Early Access". I'm okay with either.

PR #167, opened 2025-06-16
fix: import missing early access snippet for properties

dskuza:


lancesnider (dismissed):

runtimes/data-binding.mdx
```diff
@@ -773,11 +774,9 @@ Property descriptors can be inspected on a view model to discover at runtime whi
     </Tab>
 </Tabs>
 
-### Mutable Properties
+References to these properties can be retrieved by name or path. Mutable properties have getters, setters, and observe operations for their values, while other properties (except nested view models) only have observe operations. Getting or observing the value will retrieve the latest value set on that properties binding, as of the last state machine or artboard advance. Setting the value will update the value and all of its bound elements.
```
HayesGordon, line 777:
```suggestion
References to these properties can be retrieved by name or path.

Some properties are mutable and have getters, setters, and observer operations for their values. Getting or observing the value will retrieve the latest value set on that property's binding, as of the last state machine or artboard advance. Setting the value will update the value and all of its bound elements.
```

runtimes/data-binding.mdx
```diff
@@ -773,11 +774,9 @@ Property descriptors can be inspected on a view model to discover at runtime whi
...
 
-View model instances have mutable properties. References to these properties can be retrieved by name. They have get, set, and observe operations. Getting or observing the value will retrieve the latest value set on that properties binding, as of the last state machine or artboard advance. Setting the value will update the value and all of its bound elements.
-
-<Note>Trigger properties do not have a get operation - only set and observe.</Note>
+<Note>Trigger properties do not have a get or set operation - only trigger and observe.</Note>
```
HayesGordon, line 779: Maybe we don't even need this note. The API for data binding is expanding and there are more conditionals now, and each property can have a unique api

lancesnider (approved):

PR #169, opened 2025-06-19
docs(react_native): Update Expo related docs

alextoudic:
The current "Adding Rive to Expo" documentation recommends running `prebuild` and manually editing generated files, which breaks compatibility with Continuous Native Generation (CNG) and prevents future prebuilds.

This PR addresses that by:
- Updating the "Adding Rive to Expo" guide to recommend using `expo-build-properties`, which enables Expo CNG
- Linking to the official Expo documentation for generating development builds, offering more flexibility than building locally
- Editing the "Loading Rive to Expo" page to include two CNG-compatible options for using local assets: `expo-asset` and `source` with `require()`

### DIFF
--- a/runtimes/react-native/adding-rive-to-expo.mdx
+++ b/runtimes/react-native/adding-rive-to-expo.mdx
@@ -3,110 +3,117 @@ title: 'Adding Rive to Expo'
 description: 'Rive React Native Expo. '
 ---
 
-To enable your Expo application to run Rive animations you'll need to include the native Android and iOS libraries.
+To use Rive with Expo, you'll need to install the `rive-react-native` package.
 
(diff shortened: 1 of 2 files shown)


runtimes/react-native/adding-rive-to-expo.mdx
```diff
@@ -3,110 +3,115 @@ title: 'Adding Rive to Expo'
...
+        stateMachineName="avatar"
+        style={{ width: 400, height: 400 }}
+      />
+    </View>
+  );
+}
 ```
 
-### Local Assets
+> **Note:**
```
lancesnider, line 108: For notes, use the `<info>` tag.

runtimes/react-native/adding-rive-to-expo.mdx
```diff
@@ -3,110 +3,115 @@ title: 'Adding Rive to Expo'
...
-const styles = StyleSheet.create({
-  container: {
-    flex: 1,
-    backgroundColor: '#fff',
-    alignItems: 'center',
-    justifyContent: 'center',
-  },
-});
-```
+Rive for iOS requires a minimum deployment target of `14.0`.
```
lancesnider, line 37: Since this is only relevant for older setups, can we throw it in an info tag, rather than including it as a step? I think lines 35 - 41, 47 - 63 with something like: 

```
<info>
If you're using Expo SDK 51 or earlier, update the minimum deployment target to `14.0`. 


If you're using CNG, you can set the minimum iOS deployment ...
</info>
```

PR #197, opened 2025-08-25
docs(web): update codesandbox to typescript

lancesnider:
When this merges

- All Web and React examples are now using Typescript
- All Web and React examples specify which file to open (the .ts or .tsx with the Rive file)

Other stuff:
- Fix from yesterday's PR: I accidentally names some non-React files to .tsx, when they should have been .ts. 
- Remove the Centaur example

This does not include legacy examples (inputs, text, etc.), which we probably won't update.


HayesGordon (approved):
Whoop! Thank you!

PR #198, opened 2025-08-27
Add docs for JS data binding artboards

lancesnider:
Also adds pretty example links: 

CleanShot 2025-08-27 at 16 50 38@2x
CleanShot 2025-08-27 at 16 52 36@2x


lancesnider:
Instead of `Example:`, maybe say `Demo:`?

PR #202, opened 2025-09-02
Put all the code examples in one place

lancesnider:
https://github.com/user-attachments/assets/4e2fc324-af86-4a59-88e5-f508d5a6e952

### DIFF
--- a/snippets/examples-cards.jsx
+++ b/snippets/examples-cards.jsx
@@ -1,5 +1,3 @@
-import { Example } from './examples'
-
 export const ExamplesCards = ({
   examples,
   runtime,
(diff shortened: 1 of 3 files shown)


snippets/examples-cards.jsx
```diff
@@ -0,0 +1,50 @@
+export const ExamplesCards = ({
+  examples,
+  runtime,
+}) => {
+  const examplesData = {
```
lancesnider, line 5: I know this should be in its own JSON or at least its own JS file. Unfortunately, Mintlify doesn't allow you to put it anywhere but inside of the component that's going to use it.

PR #214, opened 2025-09-18
[WiP] Remove demo cards

lancesnider:
The images were working locally, but not in production. I'm currently trying to get support from Mintlify to find a way to do this. Until then, we need to remove the component. 

The React component (Demo) dynamically sets the image value: <Card key={title} title={title} img={image}>. The image URL comes from an object that looks like this: {someCard: { image: '/images/myImage.jpg}} 
 
Normally Mintlify uploads any local image link to the https://mintlify.s3/ bucket. In this case, it's trying to find the image in ./images , which doesn't exist.


lancesnider:
Closing in favor of: https://github.com/rive-app/rive-docs/pull/218

PR #227, opened 2025-09-25
docs(unity): change change low-level API to legacy API

damzobridge:
This PR updates mentions of the 'low-level API' in the Unity docs to say 'legacy API' instead. It also moves features like inputs and events to a 'Legacy' section in the Unity docs.


snippets/unity/legacy-databinding-notice.mdx
```diff
@@ -0,0 +1,8 @@
+<Warning>
+  **DEPRECATION NOTICE:** This entire page documents the legacy {subject} system.
```
HayesGordon, line 2: Nice, need to use this for the stuff I changed

HayesGordon (approved):

PR #233, opened 2025-09-29
Override Mintlify's new default iframe height

lancesnider:
Mintlify added `height: 100%` to all `iframe`'s. This overrides YouTube's default height. This makes some of our videos look squished (left), others disappear: 

CleanShot 2025-09-29 at 13 00 33@2x

This'll fix it while we're waiting for Mintlify to undo what they did.

### DIFF
--- a/custom.css
+++ b/custom.css
@@ -9,3 +9,8 @@ ul li {
 ol li {
   font-size: 14px;
 }
+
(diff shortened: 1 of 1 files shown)


jeffatrive (approved):
Nice!

PR #243, opened 2025-10-02
docs(unity): fix unity image databinding API reference

damzobridge:
This PR updates the Unity docs to use the correct `Value` setter for image properties, replacing the older/temporary `SetImage()` method.

It also adds links to the main runtime docs for data binding with lists and artboards, since those aren’t currently referenced in the Unity example (which only covers simpler property types like strings and numbers).


HayesGordon (approved):
LGTM!

PR #249, opened 2025-10-07
chore(best practices): remove power of 2 advice

lancesnider:
This was important with WebGL. As far as I can tell, it's at best a micro-optimization with WebGL2.


ErikUggeldahl (commented):
Can't approve in good conscience out of ignorance. We should find a reliable source either way.

PR #262, opened 2025-10-16
chore(framer): remove old way of adding framer

lancesnider:
I kept the old page, but removed the deprecated way of doing things. Now it just links to the plugin. 

In an upcoming PR we'll add a Site Builders section.


jeffatrive (approved):
Looks good!

PR #279, opened 2025-10-27
feat(Android): Refresh main Android docs

ErikUggeldahl:
A full refresh of the Android main page documentation:
- Adds mention of the Compose API
- Adds multiple ways of constructing a view beyond XML

Also corrects an error in Android data binding docs.

Includes a bunch of minor stylistic improvements.

### DIFF
--- a/runtimes/android/android.mdx
+++ b/runtimes/android/android.mdx
@@ -6,7 +6,7 @@ description: 'The Android runtime for Rive.'
 import NoteOnFeatureSupport from '/snippets/runtimes/rendering-feature-support.mdx'
 import { Demos } from '/snippets/demos.jsx'
 
-<NoteOnFeatureSupport/>
+<NoteOnFeatureSupport />
 
 <Demos examples={['starRating']} runtime="android" />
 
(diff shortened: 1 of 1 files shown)


runtimes/android/android.mdx
```diff
@@ -12,120 +12,205 @@ import { Demos } from '/snippets/demos.jsx'
...
+
+If you have already loaded a Rive `File` instance, you can use that to initialize the view as well. See [Caching a Rive File](../caching-a-rive-file) for more details on how and why to load Rive files.
 
 ```kotlin
+// Loads bytes on the main thread for simplicity; consider loading on a background thread for production use.
+val bytes = resources.openRawResource(R.raw.rating).use { res -> res.readBytes() }
+val riveFile = File(bytes)
+val riveView = RiveAnimationView(this)
+riveView.setRiveFile(riveFile)
+riveFile.release()
```
tylernij: Should we comment that the riveFile should be released when it's no longer needed? I.e. caching? Just a thought; not that important

PR #282, opened 2025-10-27
docs(Android): fix list instance creations

tylernij:
* Fix incorrect usage of list view model instance creations


ErikUggeldahl (dismissed):
Great, thanks for updating. Just needs the rebase. I might need to re-approve after.

tylernij:
> Great, thanks for updating. Just needs the rebase. I might need to re-approve after.

Try now! @ErikUggeldahl

PR #346, opened 2025-12-16
docs(runtime): font swapping and fallback fonts

lancesnider:
Important info about fallback fonts got accidentally moved to the deprecated section along with other text-related info. This brings that content back. Additionally: 

- Add a section about font swapping.
- Add note about fallback fonts being unsupported on web
- Rewrite the content outside the tabs (the tabs are just a copy/paste from the old file)

The localization bit needs more attention later, but this at least brings back the missing info.


HayesGordon (approved):
Thanks!

PR #355, opened 2026-01-05
fix: use string URL directly with useRiveFile

mfazekas:
Update React Native examples to pass URL strings directly to `useRiveFile` instead of using object syntax.

Passing an inline object like `useRiveFile({ uri: '...' })` creates a new object reference on each render, which can cause infinite re-renders due to useEffect dependency changes.

The correct usage is:
```js
const { riveFile } = useRiveFile('https://cdn.rive.app/animations/vehicles.riv');
```

Instead of:
```js
const { riveFile } = useRiveFile({ uri: 'https://cdn.rive.app/animations/vehicles.riv' });
```

alternatives: 
- enabling react compiler
- using useMemo

Fixes: 
- See #https://github.com/rive-app/rive-nitro-react-native/issues/99
- https://rive.app/docs/runtimes/react-native/react-native#userivefile
- https://rive.app/docs/runtimes/state-machines#react-native

### DIFF
--- a/runtimes/react-native/react-native.mdx
+++ b/runtimes/react-native/react-native.mdx
@@ -346,11 +346,11 @@ This guide documents how to get started using the Rive React Native runtime. The
     Hook for loading Rive files from a URL or local source:
 
     ```javascript
-    const { riveFile } = useRiveFile({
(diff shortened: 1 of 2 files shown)


mfazekas:
@HayesGordon pls review

PR #360, opened 2026-01-07
docs(scripting): add text path effect demo

lancesnider:
CleanShot 2026-01-07 at 10 23 56@2x
CleanShot 2026-01-07 at 10 23 47@2x

Add Laura's pretty, pretty text path effect demo.

### DIFF
--- a/scripting/protocols/path-effect-scripts.mdx
+++ b/scripting/protocols/path-effect-scripts.mdx
@@ -17,7 +17,7 @@ Path Effects are Rive scripts that modify and transform path geometry in real-ti
 <Demos
   examples={[
     "scriptingBoilPathEffect",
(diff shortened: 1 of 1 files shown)


snippets/demos.jsx
```diff
@@ -235,6 +244,15 @@ export const Demos = ({
         editor: "https://static.rive.app/revs/slot_machine.rev"
       }
     },
+    scriptingPlinko: {
+      title: "Plinko - Complete Game",
+      image: "https://static.rive.app/docs/scripting-plinko-game.png",
+      description: "Build a complete Plinko game using Layout, Node, and Path Effect scripts.",
+      links: {
+        // TODO - replace with marketplace or editor links
```
HayesGordon: Same

PR #361, opened 2026-01-08
feat(Android): Compose Docs Round 1 

ErikUggeldahl:
This is as far as I was able to make it in a day of effort. It's not complete, and arguably I shouldn't have spent as much (or any) time refactoring existing docs, but I couldn't help myself.

Covers:
- Getting started (now Compose first)
- Moves Legacy getting started to its own page
- Artboards
- Layouts
- State machines
- Updates the feature support page

### DIFF
--- a/runtimes/artboards.mdx
+++ b/runtimes/artboards.mdx
@@ -124,55 +124,58 @@ Only one artboard can be rendered at a time.
     ```
   </Tab>
 
-  <Tab title="Android (Compose)">
-    By default, the `Rive` composable will select and create the default artboard specified in the Rive editor. To specify a different artboard, you must first create an `Artboard` object from a loaded `RiveFile`.
+  <Tab title="Android">
+    <Tabs>
+      <Tab title="Compose">
+        By default, the `Rive` composable will select and create the default artboard specified in the Rive editor. To specify a different artboard, you must first create an `Artboard` object from a loaded `RiveFile`.
 
-    ### Compose
-    Artboard objects can be created in Compose using the `rememberArtboard` function, which takes a Rive file and name of the artboard.
+        ### Compose
+        Artboard objects can be created in Compose using the `rememberArtboard` function, which takes a Rive file and name of the artboard.
 
-    ```kotlin
-    val artboard = rememberArtboard(myRiveFile, "My Artboard")
-    ```
-    
-    ### Outside of Compose
-    Alternatively, you can create an artboard outside of Compose contexts. Note that with this approach you are responsible for managing the artboard's lifecycle and must eventually close it with `Artboard::close()` when no longer needed, or leveraging its [`AutoClosable`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-auto-closeable.html) interface with a `use` block.
+        ```kotlin
+        val artboard = rememberArtboard(myRiveFile, "My Artboard")
+        ```
+        
+        ### Outside of Compose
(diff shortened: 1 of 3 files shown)


runtimes/android/android.mdx
```diff
@@ -14,29 +14,29 @@ import { Demos } from '/snippets/demos.jsx'
...
-### Note on Android APIs
+### The Two Android APIs
 
 Rive for Android provides two main APIs for integrating Rive into your application.
 
-**The Legacy View-based API**
+**The New Compose API (Beta)**
 
-This API is based on Android Views and XML layouts. It is stable and widely deployed in production applications. It also supports [Compose by using AndroidView](https://github.com/rive-app/rive-android/blob/master/app/src/main/java/app/rive/runtime/example/LegacyComposeActivity.kt), though it does require some boilerplate to set up.
+This API is designed for Jetpack Compose, allowing for a more modern and declarative approach to building UIs. It is currently in beta, meaning it is still undergoing testing and some issues may be uncovered during development. We will work to address these quickly as they come up.
```
HayesGordon: Might want to say that it's ready for production usage and we recommend using it. We just encourage throrough testing, etc. etc., as it's new. We want to encourage adoption at this stage.

runtimes/android/android.mdx
```diff
@@ -96,119 +98,82 @@ Follow the steps below to get started integrating Rive into your Android app.
...
-val riveView = RiveAnimationView(this)
-riveView.setRiveFile(riveFile)
-// Release the file if you no longer need it, keep if you plan to reuse it.
-riveFile.release()
-
-setContentView(riveView)
-```
+### Adding Rive to Your Composition
+<Steps>
+  <Step title="Creating a Worker">
```
HayesGordon: Not part of this PR. But something we should document is the importance of owning this resource, deciding when to share a worker, or spin up multiple. This should be a page on its own that all the runtimes reference.

Might actually be worth adding a small `Note` or a `Warning` now. Also need to point out that a worker owns the decoded assets, so if you want to share resources you should share a single worker.

runtimes/artboards.mdx
```diff
@@ -1,24 +1,24 @@
...
-Only one artboard can be used at a time.
+Only one artboard can be rendered at a time.
 
 <Tabs>
 
   <Tab title="Web">
     ```javascript
     new rive.Rive({
-        src: 'https://cdn.rive.app/animations/vehicles.riv',
+        src: 'https://cdn.rive.app/animations/my_rive_file.riv',
```
HayesGordon: Note that `vehicles.riv` points to an actual file that exists on this CDN. So copy pasing this code would work.

runtimes/artboards.mdx
```diff
@@ -124,25 +124,55 @@ Only one artboard can be used at a time.
     ```
   </Tab>
 
-  <Tab title="Android">
-    ### Via XML
+  <Tab title="Android (Compose)">
```
HayesGordon: Same question as earlier. This is giving the impression that the new API will only ever support Compose, is that the case?

runtimes/layout.mdx
```diff
@@ -1,640 +1,658 @@
...
 
 </Tabs>
 
-## Additional Layout Options
+## Responsive Layouts
 
-If the graphic doesn’t use Rive’s Layout feature, you can configure the layout with other **Fit** options and **Alignment** settings. See the sections below for more information on **Fit** and **Alignment**.
+Rive’s layout feature lets you design resizable artboards with built-in responsive behavior, configured from the editor. Ensure the fit mode is set to **Layout** at runtime and the artboard will resize to fill its container according to the constraints defined in the editor.
+
+Optionally you may provide a **layout scale factor** to multiply the scale of the content. This allows fine tuning the visual size within your container. This property only applies when setting the **Fit** mode to **Layout**.
```
HayesGordon: Note that you removed the warning that `Aligment` has no affect when using `Fit.Layout`. Calling this out as it caused recent confusion internally as well

runtimes/artboards.mdx
ErikUggeldahl: Moved to sub-tabs.

PR #374, opened 2026-01-12
feat(Android): Update sizes to account for scripting

ErikUggeldahl:
Re-running size calculations on the new Android runtime with scripting included. The size doesn't seem to increase much, mostly because the size without scripting seems to have gone down, which I can't explain.


HayesGordon (approved):
LGTM! Thanks!

PR #389, opened 2026-01-13
feat(Android): Compose data binding docs

ErikUggeldahl:
Adds the docs for Compose data binding. Includes minor updates to legacy docs and style corrections.


ErikUggeldahl:
> Looks good. Only thing that looked out of place was the section in Auto Binding:

Anything you would change? This was a decision made a couple months ago as I was rounding out the API. Initially I tried to do the callback approach but in practice it was cumbersome and worse than this simple approach. Open to suggestions.

PR #621, opened 2026-04-14
fix(docs): Replace relative paths with absolute paths

tylernij:
Replaces the relative paths in `docs.json` with absolute paths to satisfy Mintlify's CLI.


lancesnider:
Are these currently causing a failure? If possible I'd prefer to use relative, as those will get checked for link rot and still work even if we move the page.

PR #622, opened 2026-04-14
docs: add overview to editor artboards pages

lancesnider:
This file got misnamed at some point so it was missing from the various runtime artboards pages. Here's what's being added: 

CleanShot 2026-04-14 at 11 44 55


snippets/example-eye-follow.jsx
```diff
@@ -1,5 +1,3 @@
-import { useEffect, useState } from "react";
```
lancesnider, line 1: Mintlify already includes these. This was causing a warning.

tylernij (approved):

PR #624, opened 2026-04-14
docs(editor): data binding migration guide

lancesnider:
_This is part of a larger project focused on improving the data binding docs_ 

Right now the Overview has a bunch of info about why we switched from Inputs and Events. This PR  moves that info to its own page and rewrites it to read more like a migration guide. 

It's a little longer, but there is less text overall. And I put a lot of it in lists and <Info> callouts to make it more scannable. 

This also includes instructions for automatic conversion, which didn't exist when these docs were first written.

### DIFF
--- a/editor/data-binding/overview.mdx
+++ b/editor/data-binding/overview.mdx
@@ -54,13 +54,13 @@ These exported instances can then be assigned to an artboard at runtime by devel
 
 ### Binding
 
-A binding is an association between a view model property and an editor element. For instance, you might have a string property called "name" bound to a text run. Once bound, the text run will take the value from the view model property. See [Bind Direction](#bind-direction) for additional options. 
+A binding is an association between a view model property and an editor element. For instance, you might have a string property called "name" bound to a text run. Once bound, the text run will take the value from the view model property. See [Bind Direction](#bind-direction) for additional options.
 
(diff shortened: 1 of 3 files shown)


toniarr (dismissed):

editor/data-binding/overview.mdx
```diff
@@ -98,63 +98,6 @@ A converter is a general purpose way of transforming a binding's value when it i
 
 To apply a converter on a value that already has a binding, right click on the bound property, click Update Bind, and select your converter from the Convert dropdown.
 
-# Comparing to Existing Features
-
-Data binding fills some of the same roles as existing features in Rive. In general, it is considered a more powerful alternative to both inputs and events, and we recommend you adopt it for most use cases going forward. However, this does not mean that you need to retrofit existing files as they will continue to work as expected.
-
-### Type Support
```
HayesGordon, line 105: Would we want to keep this in the migration guide?

HayesGordon (dismissed):
Nice!!

editor/data-binding/overview.mdx
HayesGordon: The type support ☝️

Tod-Rive (approved):

PR #635, opened 2026-04-24
docs: semantics docs

lancesnider:
Remaining todo's: 

- [x] Feature support (just need versions)
- [x] There's at least one page that says we don't support semantics
- [ ] Runtime flags
- [ ] Images (waiting for UI updates)


docs.json
```diff
@@ -206,6 +206,12 @@
                   "editor/layouts/scrolling"
                 ]
               },
+              {
+                "group": "Accessibility & Interaction",
```
HayesGordon: I recall us chatting about this, but I don't remember why we landed on `& Interaction`?

I'd prefer just `Accessibility`.

editor/accessibility-interaction/semantics.mdx
```diff
@@ -0,0 +1,157 @@
+---
+title: "Semantics"
+description: ""
+---
+
+Semantics describe what an element is and how it should be understood. Adding semantics makes your experiences accessible, allowing screen readers to correctly describe and navigate your content.
+
+At a high level, semantics are made up of:
+
+- [Type](#semantic-type) — what the element is (for example, button, image, or heading)
```
HayesGordon: Not `Type` but `Role`

editor/accessibility-interaction/semantics.mdx
```diff
@@ -0,0 +1,157 @@
+---
+title: "Semantics"
+description: ""
+---
+
+Semantics describe what an element is and how it should be understood. Adding semantics makes your experiences accessible, allowing screen readers to correctly describe and navigate your content.
+
+At a high level, semantics are made up of:
```
HayesGordon: We should also include **Semantic Actions** somewhere.

HayesGordon: I like this description for Roles, Traits, and States:

#### Architecture: Role / Trait / State
- Role - what the node is (button, checkbox, tab, etc.)
- Trait - what the node can do (expandable, selectable, checkable, etc.)
- State - what the node is doing (expanded, selected, checked, etc.)

Some state flags are only meaningful when their corresponding trait is set. Without the trait, the platform sees the property as not applicable.

editor/accessibility-interaction/semantics.mdx
```diff
@@ -0,0 +1,157 @@
...
+
+- [Type](#semantic-type) — what the element is (for example, button, image, or heading)
+- [Properties](#semantic-properties) — additional information like label, value, or hint
+- [Traits](#traits) — capabilities an element has (for example, selectable or expandable)
+- [State](#state) — the current value of a trait (for example, selected or disabled)
```
HayesGordon: Not all states relate to a trait. It's just that some states are gated by a trait

editor/accessibility-interaction/semantics.mdx
```diff
@@ -0,0 +1,157 @@
...
+At runtime, semantics are converted into the platform’s accessibility system.
+
+On web, this creates accessible DOM elements (for example, divs with roles and attributes). On iOS and Android, a semantic tree is generated and exposed to the system’s accessibility APIs.
+
+Rive handles mapping your semantics to each platform automatically.
+</Note>
+
+## Adding Semantics
+
+<img src="/images/accessibility-interaction/semantics/adding-semantics.gif" width="300" />
```
HayesGordon: Might not be immediately clear where to find this, as the video doesn't show it's the inspector

editor/accessibility-interaction/semantics.mdx
```diff
@@ -0,0 +1,157 @@
...
+
+<Steps>
+  <Step title="Add Semantics">
+    Select an element in your scene. In the right sidebar, click the `+` button in the **Semantics** panel to enable semantics for that element.
+  </Step>
+  <Step title="Choose a Type">
```
HayesGordon: Role not Type

editor/accessibility-interaction/semantics.mdx
```diff
@@ -0,0 +1,157 @@
...
+Available properties vary depending on the selected semantic type.
+
+### Label
+
+The label describes what the element is or does. This is what screen readers announce.
+
+For example, a slider that controls volume might have the label:
+`Volume`
+
+<Info>
```
HayesGordon: Let's remove this. This is true, but not all platforms and screen readers are equal. For example, the web might announce "expanded", but iOS does not. There might be times when authors want to provide a more unique experience by controlling the label instead of relying on the role.

editor/accessibility/semantics.mdx
```diff
@@ -0,0 +1,157 @@
...
+The label describes what the element is or does. This is what screen readers announce.
+
+For example, a slider that controls volume might have the label:
+`Volume`
+
+<Info>
+  Avoid including the type in the label (for example, "slider" or "button"). Screen readers already announce the element’s type.
+</Info>
+
+<Info>
```
HayesGordon, line 73: Let's add a new section called "Semantic Inference".

Here we can detail that certain Rive components will infer it's label (and other properties) automatically. At this time it's only a `Text` component that will automatically set its label to the text set.

editor/accessibility/semantics.mdx
```diff
@@ -0,0 +1,157 @@
...
+
+Each type has different available [properties](#semantic-properties).
+
+If your element doesn’t match a specific type, you can leave it set to **None**.
+
+<Info>
+  Some elements, like text inputs, automatically infer their semantic type.
+</Info>
+
+## Semantic Properties
```
HayesGordon, line 60: Might be nice to state the type these properties should be. At this stage it's all strings, but maybe in the future we others.

editor/accessibility-interaction/semantics.mdx
```diff
@@ -0,0 +1,157 @@
...
+
+---
+
+## State
+
+States represent the current value of a trait.
+
+For example, an element with the **selectable** trait will have a **selected** state.
+
+Use data binding to keep semantic states in sync with your Rive file. For example, bind a boolean from your view model to the **toggled** state of a switch.
```
HayesGordon: You can also key the state on an animation timeline

editor/accessibility-interaction/semantics.mdx
```diff
@@ -0,0 +1,157 @@
...
+
+---
+
+## Testing Semantics
+
+Adding semantics is only the first step. You should always test your experience at runtime to make sure it behaves as expected.
+
+Small issues—like a missing label or incorrect state—can make elements confusing or unusable for screen reader users.
+
+### How to Test
```
HayesGordon: Maybe worth calling out that this should be tested at runtime, and not the editor. Or maybe soon we can say the mirror apps

editor/accessibility/semantics.mdx
```diff
@@ -0,0 +1,157 @@
+---
+title: "Semantics"
+description: ""
+---
+
+Semantics describe what an element is and how it should be understood. Adding semantics makes your experiences accessible, allowing screen readers to correctly describe and navigate your content.
```
HayesGordon, line 12: At some point early in the document, we should highlight that this is in Early Access, and that runtime support is currently in development or released as experimental.

Inconsistencies should be reported to us, but it's worth noting that not all Roles we expose are valid across every runtime. We will document these over time.

We also need a runtime page to detail how to enable (opt-in at this stage), plus how to test etc. But that can be a separate PR.

Then we also need to update the feature support page to list out the state of this feature support on runtimes. Currently we can say that Flutter has experimental support from version `0.14.6`

tylernij (dismissed):
Gordon has some suggestions but afterwards 👍

docs.json
lancesnider: I think it was to better encompass focus, but I think "Accessibility" is good enough. :thu

editor/accessibility-interaction/semantics.mdx
lancesnider: What's the difference between a locked trait and a traitless state? What makes `enableable` a trait with a lock icon  and `obscured` not a trait? 

CleanShot 2026-04-28 at 16 00 02@2x

editor/accessibility-interaction/semantics.mdx
lancesnider: I like it. 👍

editor/accessibility-interaction/semantics.mdx
lancesnider: Never mind. Your other comment explains it " Some state flags are only meaningful when their corresponding trait is set. Without the trait, the platform sees the property as not applicable."

editor/accessibility-interaction/semantics.mdx
lancesnider: Updated. 

btw There are other gifs to add, but the ui is still WiP.

editor/accessibility/semantics.mdx
lancesnider: I'm gonna push back a little. :) 

Inference isn't a sibling section to traits and properties, it's a child of both. It's also not something you _need_ to learn about - it feels like something most people will figure out on their own. 

The only time people might look for it in the docs is if they've added a text field and can't figure out why they can't change the role. In that case, they'll go to the "Role" section, not the "semantic inference" section.

lancesnider:
Thanks @HayesGordon and @tyler. I've implemented most of your feedback, though there are a couple of todo's I'll add to the description.

runtimes/web/text.mdx
```diff
@@ -122,89 +122,4 @@ onLoad: () => {
 
 <SemanticsForAccessibility />
 
-The following code snippets provide guidance on how to add semantic labels to your Rive animations.
-
-#### Adding ARIA Label
-
-At a minimum - if it is important to convey the text value displayed in the Rive animation to all users, add an `aria-label` to the `<canvas>` element with the text value from the animation. Screen readers may read this label out immediately as it parses out the DOM contents. You'll also want to add `role="img"` to the `<canvas>` element as well.
```
zplata, line 129: nit: this small aria label section could still be useful as a "at a minimum, if semantic nodes are not used in your Rive graphic"

editor/accessibility/semantics.mdx
HayesGordon: There will be more though, `Text`, `TextInput`, maybe certain layouts as well.

Focus is also an inferred value based off whether there exists focus.

And then there is also some future complexity we might add to prune subtrees and only infer the most sensible lable. These are all things that will need to be documented. But we can figure this out later.

editor/accessibility/semantics.mdx
```diff
@@ -0,0 +1,198 @@
...
+- [Actions](#actions) - interactions triggered by the user (tap, increase, decrease)
+
+<Note>
+**Semantics at Runtime**
+
+At runtime, semantics are converted into the platform’s accessibility system.
+
+On web, this creates accessible DOM elements (for example, divs with roles and attributes). On iOS and Android, a semantic tree is generated and exposed to the system’s accessibility APIs.
+
+Rive handles mapping your semantics to each platform automatically.
```
HayesGordon: Maybe we should point to the feature support page here?

editor/accessibility/semantics.mdx
```diff
@@ -0,0 +1,198 @@
...
+</Note>
+
+## Adding Semantics
+
+![Adding semantics](/images/accessibility/semantics/adding-semantics.gif)
+
+<Steps>
```
HayesGordon, line 36: Do we want to include steps for `traits` and `states`?

editor/accessibility/semantics.mdx
```diff
@@ -0,0 +1,198 @@
...
+
+For example, an element with the **selectable** trait will have a **selected** state.
+
+Some state flags are only meaningful when their corresponding trait is set. Without the trait, the platform sees the property as not applicable.
+
+Keep semantic states in sync with your Rive file. For example, bind a boolean from your view model to the **toggled** state of a switch. You could also key the **toggled** state in a timeline.
+
+### Hidden
+
+Hides the element from the accessibility tree.
```
HayesGordon, line 122: Editor actions that remove an element visually will also prune/hide it from the accessibility tree (as well as its children). For example, toggling the visibility icon in a layout. I think that is the only one.

Opacity 0 does not remove a semantic element. If you intend a zero opacity to remove the element, then you should toggle the `hidden` state.

editor/accessibility/semantics.mdx
```diff
@@ -0,0 +1,198 @@
...
+  </Step>
+  <Step title="Listen to a Semantic Action">
+    With your listener selected, in the right sidebar choose **Semantic Action** in the **Listen to** dropdown.
+  </Step>
+  <Step title="Set the action type">
+     Select the [action type](#types-of-actions) to listen for.
+  </Step>
+</Steps>
+
+### Types of Actions
```
HayesGordon, line 157: This is not exhaustive; we will add more. And maybe we should call that out? Same for the roles/traits/states. All of it _can_ expand if there is a need.

editor/accessibility/semantics.mdx
```diff
@@ -0,0 +1,198 @@
...
+
+Actions include **tap**, **increase**, and **decrease**.
+
+#### Tap
+
+Represents activating an element (similar to a click).
+
+In most cases, **tap** is handled automatically by the platform and triggers standard pointer events. For example, activating a button with a screen reader behaves the same as a click, whether or not a tap listener is present.
```
HayesGordon: This isn't true, right now you HAVE to add a semantic tap action, otherwise the button/tap won't do anything.

There will _likely_ be a future where we just auto-add the semantic tap action for certain listeners and give you the option to remove in the editor, instead of making it a two-step process.

Let's instead document that you should ensure you have the correct listeners attached for things to work

editor/accessibility/semantics.mdx
```diff
@@ -0,0 +1,198 @@
...
+
+You typically only need a tap listener when implementing custom interactions.
+
+#### Increase / Decrease
+
+Used for adjustable elements like sliders.
+
+Users trigger these actions through assistive technologies—for example, by focusing a slider and swiping up or down with a screen reader, or using arrow keys on a keyboard.
+
+Unlike tap, these actions are not handled automatically. You should listen for **increase** and **decrease** and update your value accordingly.
```
HayesGordon: This should be remove, explained above, tap is not automatically handled.

editor/accessibility/semantics.mdx
HayesGordon: Clarification, technically, if you forget to include an action type, then tap will be invoked. But that's not something they should rely on, and instead will be the way we introduce scripted listeners later. This will then be treated as:
- no type, you then listen to it all and pass to a script.

HayesGordon (dismissed):
Looks great! Thanks for this. A few more suggestions, but those can be follow ups.

editor/accessibility/semantics.mdx
lancesnider: They're all just properties on a Role so I think I'll add state and traits to step 3: 

```
Update the element’s [semantic properties](#semantic-properties), [Traits](#traits), and [State](#state).
```

editor/accessibility/semantics.mdx
lancesnider: Great. Added: 

> Actions include **tap**, **increase**, and **decrease**. Additional actions may be added in future releases.

Also added the following to the EA note at the top: 

> Have feedback? Join the [Early Access community](https://community.rive.app/c/early-access/) to share your thoughts and help shape the feature.

editor/accessibility/semantics.mdx
lancesnider: Oh, that's great. Added a note.

runtimes/web/text.mdx
lancesnider: 👍

editor/accessibility/semantics.mdx
lancesnider: That makes sense, thanks. Added.

editor/accessibility/semantics.mdx
lancesnider: Glad you caught that, thanks!

HayesGordon (approved):
LGTM!

PR #637, opened 2026-04-28
docs(js): update available public APIs and parameters when using the high-level JS runtime

zplata:
Giving the [Rive Parameters page](https://rive.app/docs/runtimes/web/rive-parameters) a refresh on the APIs we want to surface to users that are available, clarifying some descriptions there, and pointing to relevant other pages for further guidance on usage.


lancesnider (approved):
Deeeeelightful. 👍

PR #640, opened 2026-04-29
docs(editor): document View Model Property Change as a Listener Condition

isaganttus:
Adds missing documentation for the View Model Property Change listener condition, which was untested and not reflected in the official docs.                                                                       
                                                                                                                                                                                                                    Changes:                                                                                                                                                                                                           - Added **View Model Property Change** to the Listener Condition section, explaining that it fires on any property change regardless of direction, and requires the Artboard as the Target 
- Added a screenshot example showing a boolean property being listened to                                                                                                                                        - Added an Info callout noting that since Events are deprecated, this is the recommended pattern for reacting to State Machine changes — particularly via Trigger properties as a replacement for Events
- Renamed the **User Action** section to **Listener Condition** to better reflect that it covers both pointer-based interactions and data-driven conditions


editor/state-machine/listeners.mdx
```diff
@@ -78,6 +78,16 @@ Available actions include:
...
 
+**View Model Property Change** – Triggers when a selected View Model property changes. Only available when the Artboard is set as the Target. Use the dropdown to select the specific property to listen to.
+
+Note that it is not possible to specify an expected value—the listener fires whenever the property changes, regardless of the direction of the change. For example, a Boolean listener will fire whether the value changes from `true` to `false` or from `false` to `true`.
+
+![Listener configured to listen for a boolean View Model property change, with the Artboard set as the Target](/images/editor/state-machine/listener-view-model-property-change-boolean.png)
+
+<Info>
+  Since Events are deprecated, **View Model Property Change** is the recommended way to react to changes happening inside a State Machine. In particular, Triggers are a direct replacement for Events: a Trigger can be fired from within the State Machine (e.g. on a state transition), and a listener set to **View Model Property Change** on that Trigger property will fire in response—giving you the same notification pattern without relying on Events.
```
lancesnider:
```suggestion
  Since General Events are deprecated, **View Model Property Change** is the recommended way to react to changes happening inside a State Machine. In particular, Triggers are a direct replacement for Events: a Trigger can be fired from within the State Machine (e.g. on a state transition), and a listener set to **View Model Property Change** on that Trigger property will fire in response—giving you the same notification pattern without relying on Events.
```

lancesnider (approved):
This is great. Thanks Isa!

PR #643, opened 2026-04-30
docs(editor): move inputs, runtime events to legacy features

lancesnider:
1. Move `Inputs` out of the main nav, into the `Legacy Features` dropdown at the bottom of the sidebar. 
2. The `Events` page says Events are deprecated, which they are not. 
3. Split `Events` into 2 pages - one for valid events, one for the legacy `Events at runtime`

The `Events` page still needs a lot of love, but this at least gets rid of incorrect info.


editor/events/overview.mdx
```diff
@@ -1,23 +1,10 @@
...
+description: "Creating Rive Events"
 ---
 import { YouTube } from "/snippets/youtube.mdx";
 
-<Warning>
-  **DEPRECATION NOTICE:** This entire page documents the legacy Events system.
-  **For new projects:** Use [Data Binding](/runtimes/data-binding) instead.
-  **For existing projects:** Plan to migrate from Events to Data Binding as soon
-  as possible. **This content is provided for legacy support only.**
-</Warning>
```
lancesnider, line 13: This isn't true. Events are still valid, just not for communicating with the runtime code.

editor/events/events-at-runtime.mdx
```diff
@@ -0,0 +1,38 @@
...
+  **This content is provided for legacy support only.**
+</Warning>
+
+For general information about adding, creating, and signaling events, see [Rive Events](/editor/events/overview).
+
+<YouTube id="M5DIDVtYI_Y" />
+
+Events are a (deprecated) way to send signals to your runtime code to execute a block of code at the right moment. They enhance the communication between designers and developers by passing along useful information. With them, we can do things like go to a URL, play a sound, have some HTML appear, or do anything else we may want to accomplish via code.
+
+Coordinating Rive events at design time and runtime will be important to ensure a successful integration in apps, games, and more.
```
ErikUggeldahl: Nit, grammar: "Coordinating Rive events between design and runtime is important to ensure"

editor/events/events-at-runtime.mdx
```diff
@@ -0,0 +1,38 @@
...
+  **For new projects:** Use [Data Binding](/runtimes/data-binding) instead.
+
+  **This content is provided for legacy support only.**
+</Warning>
+
+For general information about adding, creating, and signaling events, see [Rive Events](/editor/events/overview).
+
+<YouTube id="M5DIDVtYI_Y" />
+
+Events are a (deprecated) way to send signals to your runtime code to execute a block of code at the right moment. They enhance the communication between designers and developers by passing along useful information. With them, we can do things like go to a URL, play a sound, have some HTML appear, or do anything else we may want to accomplish via code.
```
ErikUggeldahl: Nit: "right moment" is a bit vague. I would clarify as something like "to execute a callback synchronized to a moment on a timeline." or something like that.

editor/events/events-at-runtime.mdx
```diff
@@ -0,0 +1,38 @@
...
+
+Coordinating Rive events at design time and runtime will be important to ensure a successful integration in apps, games, and more.
+
+### Properties (Deprecated)
```
ErikUggeldahl: Is it necessary to mark these as deprecated when the whole runtime side is deprecated?

editor/events/events-at-runtime.mdx
```diff
@@ -0,0 +1,38 @@
...
+
+Properties allow us to define extra information being passed to the Runtimes. For example, you may want to pass in the name of an audio file to play when the event is reported at runtime, or perhaps some other metadata for data analytics purposes.
+
+To add a new property, hit the plus button next to the Properties.
+
+![Add New Property](https://ucarecdn.com/d4cd3b8f-3765-4c28-9204-e5daf7fff0d8/)
+
+First, we want to change the name of our property to something identifiable. Next, we need to select what type of value our property will track such as a Number, Boolean, or String.
```
ErikUggeldahl: "such as" doesn't read as exhaustive. Could strengthen to: "which can be one of ...".

Also, "track" seems like the wrong verb. "send"?

editor/events/events-at-runtime.mdx
```diff
@@ -0,0 +1,38 @@
...
+
+![Rename and select input](https://ucarecdn.com/73d05fb1-7c9c-4c9c-a17c-51781ef30d0e/)
+
+You can key a property to let the runtimes know that a particular boolean, number, or string property has a new value.
```
ErikUggeldahl: This could use improvement. Does keying the property value automatically send the event? I wouldn't think so, but that's how this reads.

editor/events/overview.mdx
```diff
@@ -1,23 +1,10 @@
...
-
-
-<YouTube id="M5DIDVtYI_Y" />
-
-Events are a way to send signals to your runtime code to execute a block of code at the right moment. They enhance the communication between designers and developers by passing along useful information. With them, we can do things like go to a URL, play a sound, have some HTML appear, or do anything else we may want to accomplish via code.
```
ErikUggeldahl, line 18: Without this, we no longer have a high level explanation of what events are good for. We now jump straight to "Creating an Event". With events not being deprecated on the editor side (which I don't fully understand - why can't data binding do all of what they can?), we should provide a new explanation for their purpose.

editor/events/overview.mdx
```diff
@@ -35,7 +22,7 @@ Once we've added an event, we need to configure the Event using the Inspector.
 
 ### Name
 
-The Name field is where we can give our Event a specific name. It's important to do this so that at runtime, we can hook up the correct bits of code to their corresponding events.
+The Name field is where we can give our Event a specific name.
```
ErikUggeldahl: A bit tautological. Anything we can provide to make this statement a bit more useful? E.g. useful for describing when or why the event should fire, or what its firing represents.

editor/events/overview.mdx
```diff
@@ -47,19 +34,8 @@ The Type dropdown allows you to change the Event type between General and URL.
...
-To add a new property, hit the plus button next to the Properties.
-
-![Add New Property](https://ucarecdn.com/d4cd3b8f-3765-4c28-9204-e5daf7fff0d8/)
-
-First, we want to change the name of our property to something identifiable. Next, we need to select what type of value our property will track such as a Number, Boolean, or String.
-
-![Rename and select input](https://ucarecdn.com/73d05fb1-7c9c-4c9c-a17c-51781ef30d0e/)
-
-### URL
+### URL Properties
```
ErikUggeldahl: Are we sure we want to continue supporting URL events? This is only for runtime integration, and that integration is deprecated.

editor/events/overview.mdx
ErikUggeldahl: Shouldn't Audio be listed here?

editor/events/overview.mdx
```diff
@@ -47,19 +34,8 @@ The Type dropdown allows you to change the Event type between General and URL.
...
 
 When an `Open URL` Event is selected, we have additional configuration options.
```
ErikUggeldahl: "`Open URL` Event" should maybe add the word "type" to the end.

editor/events/overview.mdx
```diff
@@ -80,6 +56,18 @@ The Target tells the user's browser where this URL should open. We have a few op
 
 <Info>At this time, by default, Rive will not open URLs when this type of event is reported in embed URLs or Marketplace posts due to security considerations. However, this may change in the future</Info>
 
+### Audio Properties
+
+See [Audio Events](/editor/events/audio-events).
+
+### Properties (Deprecated)
+
+Using events to communicate with runtime code is deprecated. When this was still valid, properties were a way to add values to Events.
```
ErikUggeldahl: "When this was still valid, properties were a way to add values to Events."

Phrasing is a bit awkward. Maybe we can use present tense, noting that they are still deprecated.

"Properties allow an event to carry information to runtime. See [Events at Runtime]. This feature is deprecated and is now accomplished with [Data Binding]."

Something like that.

editor/events/overview.mdx
```diff
@@ -80,6 +56,18 @@ The Target tells the user's browser where this URL should open. We have a few op
...
+
+To communicate with runtime code, [Data Binding](/editor/data-binding) instead.
```
ErikUggeldahl: If you don't use the above phrasing, this is missing a word, either "use" or "see" [Data Binding].

editor/events/overview.mdx
```diff
@@ -80,6 +56,18 @@ The Target tells the user's browser where this URL should open. We have a few op
...
+
+---
```
ErikUggeldahl: Is there a reason for the rule break here? Isn't that transition to a new topic mostly denoted by the change in header?

editor/events/overview.mdx
```diff
@@ -92,18 +80,12 @@ First, select the timeline you want to add the event to. Next, use the Report Ev
 
 ![Keying an Event on the timeline](https://ucarecdn.com/bd8d36f9-9cd1-4eec-9c37-85d4a0a19643/)
 
-Additionally, you can key a property to let the runtimes know that a particular boolean, number, or string property has a new value.
-
 ### Transition & State
 
-You can report an event on a Transition or a State. We typically do this to signal at runtime contextual information about what is happening in the State Machine. For example, if we want to have some element appear at the end of our Transition, we would want to use an event tied to a Transition to signal this.
+You can report an event on a Transition or a State.
```
ErikUggeldahl: This made more sense as an independent paragraph when it carried the other information. I'd now tie this sentence to the next paragraph to avoid it feeling isolated.

ErikUggeldahl (changes requested):
Some minor things to consider, mostly around clarity! Thanks for doing this.

editor/events/overview.mdx
lancesnider: I'm glad you asked because this is something I'm slightly torn on. I feel like our h2's and h3's are similar enough that I don't always notice when I've moved to a top-level section. I've been adding these a little at a time and am planning a PR to do them all at once. 

Agree, disagree? 
CleanShot 2026-04-30 at 11 03 37

editor/events/overview.mdx
lancesnider: I was really struggling with this sentence, thanks!

editor/events/overview.mdx
lancesnider: URL events kind of suck because I believe there are permissions issues. I'll find out if they are all the way deprecated.

editor/events/overview.mdx
lancesnider: Yeah, that's not great, is it. Changing it to: 

Give your event a name so it’s easy to identify and reference.

You can rename it using the **Name** field, or by double-clicking the name directly on the artboard.

editor/events/events-at-runtime.mdx
lancesnider: Yes, that's exactly how it works. It's just like a trigger, but triggers can't be keyed.

editor/events/events-at-runtime.mdx
lancesnider: It's annoying to repeat stuff, but it's for people who link directly to properties, rather than starting at the top of the page. I've also read that it's important for AI, which doesn't always understand that subsections are also deprecated.

editor/events/overview.mdx
lancesnider: There's really only one remaining use. You can't key a view model property or trigger, but you can key an event which can affect vm props. It's kind of roundabout. :/

editor/events/events-at-runtime.mdx
lancesnider: I looked into this more and no, we don't need the extra text. Removing. 👍

lancesnider:
@ErikUggeldahl I chatted with Hernan and Robert and ended up redoing a bunch of this. Most importantly, we're deprecating general events. 🎉

editor/events/open-url-events.mdx
```diff
@@ -0,0 +1,37 @@
+---
+title: 'Open URL Events'
+description: "Using events to open URLs at runtime"
+noindex: true;
+---
+
+For general information about adding, creating, and signaling events, see [Rive Events](/editor/events/overview).
+
+### URL Properties
```
ErikUggeldahl, line 23: As a note, this functionality was never added to the command queue, and so we have no way of responding to them. We need to consider as a team whether we want this functionality. Is it a security risk to allow Rive files to open arbitrary URLs? What if they point a user to a compromised website or a phishing scam?

ErikUggeldahl: This was always more of a convenience as well. Developers could still react to a data binding event and open a browser using whatever native APIs are available to do that.

editor/events/overview.mdx
```diff
@@ -1,109 +1,71 @@
...
-### Type
+<Steps>
+  <Step title="Create an Event">
+    Use the Events tool located in the Toolbar and click anywhere on the artboard.
 
-The Type dropdown allows you to change the Event type between General and URL.
+    ![Adding a new event](https://ucarecdn.com/4ed6c563-4c59-42c8-b40c-f502d5a8e1a4/)
 
-![Image](https://ucarecdn.com/9621c007-de2e-428c-95d7-837615a37caa/)
+    You'll notice that the Event is displayed on the artboard and in the Hierarchy.
```
ErikUggeldahl: Nit: mixed casing on "Event". Here you capitalize it, next step it's lower.

editor/events/overview.mdx
```diff
@@ -1,109 +1,71 @@
...
+    Give your event a name so it’s easy to identify and reference.
 
-### Properties
+    You can rename it using the **Name** field, or by double-clicking the name directly on the artboard.
 
-Properties allow us to define extra information being passed to the Runtimes. For example, you may want to pass in the name of an audio file to play when the event is reported at runtime, or perhaps some other metadata for data analytics purposes.
+    ![Renaming an Event](https://ucarecdn.com/4558fb61-4649-4210-9ec6-c828c48ab2b2/)
+  </Step>
+  <Step title="Select an Event type">
+    The Type dropdown allows you to change the Event type between Audio, URL, and General.
```
ErikUggeldahl: Nit: Here and below, I'd keep consistent casing with "Event type" - either both lower case or both upper as one proper noun.

editor/events/overview.mdx
```diff
@@ -1,109 +1,71 @@
...
-- `Parent` - Opens in the parent browsing context of the current one. If no parent, behaves as `Self`.
-- Self - Opens in the current browsing context
-- Top - Opens in the topmost browsing context (the "highest" context that's an ancestor of the current one). If no ancestors, behaves as `Self`.
-
-<Info>At this time, by default, Rive will not open URLs when this type of event is reported in embed URLs or Marketplace posts due to security considerations. However, this may change in the future</Info>
 
 ## Signaling an Event
 
-We can signal an event in three ways: via the timeline, on a state, or on a transition.
+We can signal an Event in four ways: from a timeline, a listener, a state, or a transition.
```
ErikUggeldahl: Nit: More casing on "Event" - do we want this to be capitalized everywhere? E.g. we don't capitalize artboard, which is very much a Rive proper term.

ErikUggeldahl (dismissed):
Nits on casing and a comment about whether we want to allow URLs in the future. + perhaps we want to note that they won't be supported in new Android and Apple runtimes?

editor/events/overview.mdx
lancesnider: Good call. I think I have a better rule to follow now. If it's a concept, it's lowercase. If it's referring to a specific too, it's title case. 👍

editor/events/open-url-events.mdx
lancesnider: For now I'm adding a warning about Apple/Android, but  I can definitely see getting rid of these altogether. 

They tend to confuse people at runtime because they're disabled by default and aren't supported in the marketplace or share links. You need to set property to make them work.

lancesnider: 3 big new warnings that scream "maybe don't use this". 😂 

CleanShot 2026-05-08 at 14 13 54

RobertHaynie (approved):

PR #670, opened 2026-05-11
docs(editor): transition and state actions

lancesnider:
You can now fire actions, like setting a vm property, when states and transitions start and end.


jeffatrive (approved):
Love it

PR #702, opened 2026-05-29
docs(editor): binding data

lancesnider:
- Add "Binding Data" page
- Add UseCase snippet for written examples (It's basically just in Info callout with a lightbulb icon)


editor/data-binding/binding-data.mdx
```diff
@@ -0,0 +1,148 @@
+---
+title: "Binding Data"
+description: "Binding is how you connect data to properties and elements in your Rive file."
+---
+
+import { YouTube } from '/snippets/youtube.mdx'
+import { UseCase } from '/snippets/use-case.mdx'
+
+For example, if you have a `teamName` property bound to a text run in your scene, changing the `teamName` value updates the displayed text automatically.
```
jeffatrive, line 9: Starting with "For example" feels like something is missing. Maybe an intro sentence or two?

jeffatrive (dismissed):
One suggested edit, but nothing to prevent publishing.

editor/data-binding/binding-data.mdx
lancesnider: This is something I kinda hate about mintlify. The meta description and the subheading are the same thing. And the subheading is styled to where it's basically just the first sentence of the page. 

CleanShot 2026-06-02 at 13 16 20@2x

I'm gonna merge as is because this is a pattern we've been following for a while, but I'm considering other options.

jeffatrive (approved):

PR #719, opened 2026-06-05
docs(js): adding new focus-related params for instancing Rive for focus management

zplata:
Adding new focus-related params for instancing `new Rive()`


HayesGordon (approved):
LGTM!

PR #749, opened 2026-06-22
chore: remove unused images

lancesnider:
Get rid of 200 unused images, making it easier to keep valid images up to date.


zplata (approved):
Changes sgtm. Something might be broken with our mintlify preview though - looks like its 404'ing on the latest staging deployment

PR #759, opened 2026-06-25
docs(accessiblity): reduced mototion

lancesnider:
- Explain why we can't automatically reduce motion
- Explain reduced motion strategies
- Control `prefersReducedMotion` at runtime


snippets/demos.jsx
```diff
@@ -9,6 +9,15 @@ export const Demos = ({
   childrenIndex = 0
 }) => {
   const examplesData = {
+    accessibilityReducedMotion: {
+      title: 'Reduced Motion Example',
+      description: 'Strategies for respecting a user\'s reduced motion preferences.',
+      riv: 'https://static.rive.app/rivs/accessibility_reduced_motion.riv',
```
HayesGordon, line 17: I think you set the wrong artboard in this share link

HayesGordon (approved):
After the discussion on Monday, there should be a future where reduced motion is a first-class consideration in the Editor. At a minimum, this will mean you don't need to manually define a view model property, and might even mean we expose more edit time considerations to make building for reduced motion easier. I recall you saying you don't want to write the docs with a future-looking perspective. But maybe it's worth noting that this is the _current_ recommended way to do it?

PR #761, opened 2026-06-25
document path markn-needs-update

lancesnider:
If the path script is the only thing happening in a nested component, it'll appear to freeze.


HayesGordon (approved):
LGTM!

PR #768, opened 2026-06-29
docs(editor): lottie

lancesnider:
_This is part of a larger initiative to modernize the Assets section._ 

Moves Lottie docs from the main content into its own page. We're not going to include this page in the main nav - users will only get to it from relevant links or search. 

The main focus of this page isn't how to use lottie or its import, but to explain why it's an enterprise-only feature and to make it clear that it's a starting point, not a perfect 1:1 conversion.


lancesnider:
Closing in favor of #772

PR #793, opened 2026-07-08
docs(scripting): shaders

lancesnider:
Adds documentation for shaders and script settings.


docs.json
```diff
@@ -296,6 +296,7 @@
                   "scripting/protocols/path-effect-scripts",
                   "scripting/protocols/transition-condition-scripts",
                   "scripting/protocols/listener-action-scripts",
+                  "scripting/wgsl-shaders",
```
lancesnider, line 299: @luigi-rosso I put this under `Protocols (Script Types)`, but shaders aren't a protocol. I'm wondering if I should put shaders, util scripts, and tests somewhere else.

jeffatrive (approved):
Looks good. I tested the scripts in Early Access too and they worked.

PR #833, opened 2026-08-03
docs(scripting): inputs

lancesnider:
- Update examples and add links to marketplace items
- Refactor Inputs and Data Binding so there's no overlap
- Add a video


jeffatrive (approved):
Looks great

PR #851, opened 2026-08-11
docs(js/react): add documentation for Web (JS) and React on semantic support

zplata:
- Adds to the feature support table
- Adds docs for Semantics (new page) on Web (JS) + React sections
- Centralizes a few common doc notes from the apple runtimes into snippets


runtimes/web/semantics.mdx
```diff
@@ -0,0 +1,161 @@
...
+    riveCanvasLabel: "Login Experience",
+  },
+});
+```
+
+| Option | Type | Description |
+| --- | --- | --- |
+| `riveCanvasLabel` | `string` | `aria-label` applied to the overlay container. Defaults to `"Rive animation"`. |
+
+## How Semantics Map to the DOM
```
HayesGordon, line 97: Nice, we may need to mark some of these as Early Access vs Prod (if we make that distinction before releasing to prod)

runtimes/web/semantics.mdx
```diff
@@ -0,0 +1,161 @@
...
+
+<Note>
+  An image with no label is treated as decorative and hidden from assistive technologies, since a `role="img"` element without an accessible name is a WCAG failure. Add a label in the editor to any image that carries meaning.
+</Note>
+
+## Keyboard and Actions
+
+The overlay wires keyboard interaction to the [semantic actions](/editor/accessibility/semantics#actions) you listen for in the editor. Add a **Semantic Action** listener in your state machine to respond to them.
+
+| Browser Interaction | Semantic Action fired |
```
HayesGordon: Interesting. Are these actions you listen to manually? Or do you mean you get them out of the box through the system screen reader?

HayesGordon (dismissed):
LGTM!

runtimes/web/semantics.mdx
zplata: Out of the box - no user config

runtimes/web/semantics.mdx
zplata: @HayesGordon How about for now, we add another warning note here that some roles/traits may change while semantics is in early access
> This set of roles and traits may be subject to change while the semantic feature is in Early Access

runtimes/web/semantics.mdx
HayesGordon: Not sure if we should document it then, because the experience is different depending on platform/browser/mobile (i believe)

runtimes/web/semantics.mdx
HayesGordon: Yes, definitely!

lancesnider (dismissed):

runtimes/web/semantics.mdx
zplata: Fair. We can see if it comes up and then add clarity around the guts of what fires the semantic actions

lancesnider (approved):

PR #855, opened 2026-08-13
docs(editor): move scrolling docs from layouts to constraints

lancesnider:
Not sure why scrolling ever lived in Layouts, since it is a constraint. 

Please don't review the content of the scroll-constraint page. It's a direct copy from the old layouts/scrolling.mdx page, which I've deleted. The content isn't perfect, but it's not the focus of this PR. This PR is only about moving where the content of that page lives.


editor/constraints/scroll-constraint.mdx
```diff
@@ -2,8 +2,91 @@
 title: 'Scroll Constraints'
```
lancesnider, line 2: Please don't review the content of this page. It's a direct copy from the `layouts/scrolling.mdx` page, which I've deleted. 

This content isn't perfect, but it's not the focus of this PR. This PR is just about moving where it lives.

RobertHaynie (approved):

PR #882, opened 2026-08-23
docs(scripting): list the permitted values on enum reference pages

neurowave:
Twelve string-union types have reference pages carrying the description but not the values. These types accept nothing except those exact strings, so the pages don't say what to pass.

`BlendFactor` is the clearest case. The whole page:

```
Blend factor for src/dst color and alpha (WebGPU canonical strings).
```

13 permitted values, none listed. Same for `BlendOp`, `BufferUsage`, `CompareFunction`, `CullMode`, `Filter`, `GamepadMappingKind`, `PrimitiveTopology`, `TextureType`, `VertexFormat`, and `WrapMode`. `TextureAspect` mentions `depth-only` in prose but never lists `all` or `stencil-only`.

Values and order come from the Luau declarations in `scripting_workspace` builtins, in the bullet style already used by `StrokeCap`, `StrokeJoin`, and `PaintStyle`.

Descriptions are untouched. `GamepadMappingKind` still has none, matching the absent doc comment upstream.

`mint broken-links` passes.

### DIFF
--- a/scripting/api-reference/artboards/gamepad-mapping-kind.mdx
+++ b/scripting/api-reference/artboards/gamepad-mapping-kind.mdx
@@ -2,3 +2,5 @@
 title: GamepadMappingKind
 ---
 
+* `standard`
+* `unknown`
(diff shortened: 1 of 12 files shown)


lancesnider:
I like this idea, but these pages are auto-generated from the rive repo. Rather than adding the options manually, I'll add this to the script. That way it'll work with future enums.

lancesnider:
Here's the PR that does the same thing, but adds a few more: https://github.com/rive-app/rive/pull/13740

PR #883, opened 2026-08-23
docs(scripting): match protocol names to the Editor and list the missing ones

neurowave:
Checked the protocols overview and each protocol page against `ScriptProtocol` and the labels the type picker renders.

### Names that don't match the picker

| Docs say | Picker shows |
|---|---|
| **Util** | **Blank** |
| **Listener Action Script** | **Listener Action** |
| **Transition Condition Script** | **Transition Condition** |

"Util" doesn't appear anywhere in the editor. The label comes from `ScriptProtocol.utility.label`, which is `'Blank'`, so someone following the page looks for a menu entry that isn't there.

### Missing from the overview list

The list had 7 of the 10 protocols available today:

- **Blank** — already has a page and a sidebar entry, just not in this list
- **File Format** — undocumented
- **Text File Format** — undocumented

Both file format protocols get a one-line entry rather than a page, since neither has one yet. Verified user-reachable: the create menu renders every `availableProtocols()` entry by `type.label`.

Interpolator is excluded, still behind `ENABLE_SCRIPTED_INTERPOLATORS` (`prod: enabledForEveryone: false`). `availableProtocols()` filters exactly that one flag and nothing else.

`mint broken-links` passes.

### DIFF
--- a/scripting/debugging/unit-testing.mdx
+++ b/scripting/debugging/unit-testing.mdx
@@ -1,10 +1,10 @@
 ---
 title: "Unit Testing"
-description: "Write and run unit tests for your Blank Scripts using Test scripts."
+description: "Write and run unit tests for your helper functions using Test scripts."
 ---
 import { Demos } from '/snippets/demos.jsx'
 
-Test scripts let you write unit tests for your [Blank Scripts](/scripting/protocols/util-scripts) and run them directly in the Rive editor.
+Test scripts let you write unit tests for your helper functions (see [Blank Scripts](/scripting/protocols/util-scripts)) and run them directly in the Rive editor.
 Use them to verify math helpers, string utilities, or any other pure logic your scripts depend on.
 
 Beyond validating your code, tests also serve as precise instructions for the [AI Agent](/editor/ai-agent), helping it produce code that behaves exactly as you intend.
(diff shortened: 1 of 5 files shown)


scripting/debugging/unit-testing.mdx
```diff
@@ -1,10 +1,10 @@
 ---
 title: "Unit Testing"
-description: "Write and run unit tests for your Util Scripts using Test scripts."
+description: "Write and run unit tests for your Blank Scripts using Test scripts."
```
lancesnider: This doesn't really make sense. You don't write test for blank scripts.

scripting/protocols/listener-action-scripts.mdx
```diff
@@ -10,7 +10,7 @@ Use Listener Action Scripts when you need to perform side effects—such as upda
 
 ## Creating a Listener Action Script
 
-[Create a new script](/scripting/creating-scripts) and select **Listener Action Script** as the type.
```
lancesnider, line 13: In the editor they are called Listener Action Scripts

lancesnider (changes requested):

lancesnider (approved):

PR #884, opened 2026-08-23
docs(scripting): add VertexBuffer and TriangleBuffer reference pages

neurowave:
`Renderer.drawImageMesh` takes `VertexBuffer` and `TriangleBuffer`:

```lua
drawImageMesh(image: Image, sampler: ImageSampler, vertices: VertexBuffer, uvs: VertexBuffer, triangles: TriangleBuffer, blendMode: BlendMode, opacity: number) -> ()
```

Neither type had a page, so the signature named types you couldn't look up. Every other Renderer method links its parameter types: `drawPath` links Path and Paint, `drawImage` links Image, ImageSampler, and BlendMode.

Adds a page for each with their `add` and `reset` methods, a Mesh group in the sidebar, and links from the `drawImageMesh` description.

Content comes from the Luau declarations in `scripting_workspace` builtins. Page structure follows the existing reference pages: `## Methods`, the `{/* */}` signature comment above each signature block, and a code block rather than constructor prose, matching ImageSampler, which is constructed the same way.

`mint broken-links` passes.

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -485,6 +485,13 @@
                 ]
               },
               {
+                "group": "Mesh",
+                "pages": [
+                  "scripting/api-reference/mesh/vertex-buffer",
+                  "scripting/api-reference/mesh/triangle-buffer"
+                ]
+              },
+              {
                 "group": "Paint",
                 "pages": [
                   "scripting/api-reference/paint/paint",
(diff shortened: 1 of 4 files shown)


lancesnider:
Closing in favor of: https://github.com/rive-app/rive/pull/13740

PR #885, opened 2026-08-23
docs(scripting): document Context.log

neurowave:
The Context page documents 14 methods but not `log`, which is how a script produces output: the CLI watch log and problems surface, and the editor console.

It comes first in the `Context` declaration, so it's added in that position, with the description and example taken from the source doc comment.

`mint broken-links` passes.


lancesnider:
This is already documented.

PR #887, opened 2026-08-23
docs(runtimes): route hardcoded runtime versions through docs.json variables

neurowave:
`web-js.mdx` already reads its version from a `docs.json` variable. Three other pages pin versions inline, and all three have drifted:

| page | pinned | current |
|---|---|---|
| `runtimes/web/inputs.mdx` | `@rive-app/canvas@2.10.3` | 2.40.1 |
| `runtimes/web/low-level-api-usage.mdx` (4 places) | `@rive-app/canvas-advanced@2.26.1` | 2.40.1 |
| `runtimes/apple/apple.mdx` | `rive-ios from: "6.13.0"` | 6.24.0 |

`versionWebGL2` was also behind at 2.38.3, so that's bumped too.

The low-level page's versions are illustrative — the surrounding note says to match whatever you installed — but they're still numbers a reader copies, and a variable keeps them current for free. Adds `versionCanvas`, `versionCanvasAdvanced`, and `versionApple`, so future bumps are one edit in `docs.json` instead of a grep.

Variables already substitute inside fenced code blocks on the live site (`web-js.mdx` and the Apple pages do this today), so the `swift` and `javascript` samples render correctly.

Not touched: `runtimes/react/preloading-wasm.mdx` pairs `canvas@1.0.95` with `react-canvas@3.0.33` to illustrate version pairing, so a live number would defeat the point. Unity's `#v0.0.0` is a deliberate placeholder with "replace 0.0.0 with the latest release" beside it.

`mint broken-links` passes.


docs.json
```diff
@@ -27,7 +27,10 @@
     "pricingVoyagerYearly": "$304",
     "pricingEnterpriseYearly": "$1,440",
     "versionMajorWebGL2": "2",
-    "versionWebGL2": "2.41.0",
+    "versionWebGL2": "2.40.1",
```
HayesGordon: we can have a single web version, these will always be the same

docs.json
```diff
@@ -27,7 +27,10 @@
...
+    "versionCanvas": "2.40.1",
+    "versionCanvasAdvanced": "2.40.1",
+    "versionApple": "6.24.0",
```
HayesGordon, line 31: We should add all the runtime as part of this change?

HayesGordon: Can see this PR where I also added all the tags, as part of an effort to automate this, but parked that PR: https://github.com/rive-app/rive-docs/pull/800/changes#diff-b7b4169fa5b2e02aa17d40553a73c5d9520086753c501eeb926c63b162bb6d9e

runtimes/flutter/flutter.mdx
```diff
@@ -43,7 +43,7 @@ Follow the steps below to integrate Rive into your Flutter apps.
     ```yaml
     # pubspec.yaml
     dependencies:
-      rive: ^0.14.0-dev.6 # or latest dev version
+      rive: ^{{versionDevFlutter}}
```
HayesGordon: Lets point to stable. There are breaking changes that have not been documented

runtimes/flutter/flutter.mdx
```diff
@@ -14,7 +14,7 @@ This guide documents how to use the Rive Flutter runtime to easily integrate Riv
 
 <Note>
```
HayesGordon, line 15: Let's reword this whole note. This was a not for the last major.

We can just say:
"There is a new major version of Rive Flutter, 0.15.x`. Currently available as a dev release. Try it out and let us know what you think."

Or something like that. We can clean this up when we add the migration guide and point to it.

docs.json
```diff
@@ -27,7 +27,14 @@
     "pricingVoyagerYearly": "$304",
     "pricingEnterpriseYearly": "$1,440",
     "versionMajorWebGL2": "2",
-    "versionWebGL2": "2.41.0",
+    "versionWeb": "2.40.1",
+    "versionApple": "6.24.0",
+    "versionAndroid": "11.9.2",
+    "versionReactNative": "0.4.20",
+    "versionFlutter": "0.14.11",
+    "versionDevFlutter": "0.15.0-dev.1",
```
HayesGordon: Don't think it's worth adding `versionDevFlutter`, this will not exist that long. And ideally we should be pointing to stable versions, and we can just mention there is a new dev release. The previous time I didn't do that because that was a much harder, forced migration + overall docs change

docs.json
```diff
@@ -27,7 +27,14 @@
     "pricingVoyagerYearly": "$304",
     "pricingEnterpriseYearly": "$1,440",
     "versionMajorWebGL2": "2",
-    "versionWebGL2": "2.41.0",
+    "versionWeb": "2.40.1",
+    "versionApple": "6.24.0",
+    "versionAndroid": "11.9.2",
+    "versionReactNative": "0.4.20",
```
HayesGordon, line 33: Now that we have all of these flags, I would have imagined there would be more places in the docs that can take these now? Maybe an opportunity to add these version information for each runtime.

HayesGordon (approved):
LGTM!

PR #897, opened 2026-08-26
docs(js/react): add deprecation docs for various parameters in the high-level JS/React runtime and use new stateMachine param in examples

zplata:
Starting in `v2.41.0`/`v4.33.0` of the JS/React runtimes respectively, we added deprecations for the following:
- `animations` - initialization param
- `stateMachines` - initialization param
- `automaticallyHandleEvents` - initialization param related to Rive events
- `useStateMachineInput` hook

Adds:
- `stateMachine` - new param that is just a single string and will replace the `stateMachines` param

Those changes also deprecate events/text run/state machine input APIs so those should be touched on in the parameters pages as well.

### DIFF
--- a/runtimes/web/rive-parameters.mdx
+++ b/runtimes/web/rive-parameters.mdx
@@ -3,6 +3,8 @@ title: 'Rive Parameters'
 description: 'API docs for the Rive instance.'
 ---
 
+import DefaultPlayback from "/snippets/runtimes/state-machines/default-playback.mdx"
+
 This page is a **reference** for constructor options, types, and instance methods on the high-level **`Rive`** class, as well as a number of other exported classes. For walkthroughs and samples, use the guides below and link back here when you need exact signatures.
 
 **Related guides**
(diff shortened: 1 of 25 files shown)


runtimes/web/rive-parameters.mdx
```diff
@@ -448,6 +473,14 @@ Prefer data binding with [Text data binding](/runtimes/web/data-binding#reading-
 
 ---
 
+#### scrub()
+
+`scrub(animationNames?: string | string[], value?: number): void`
+
+Scrubs the specified timeline animations to the given time; if none are specified, scrubs all of them. Deprecated in `v2.41.0` — use a state machine to control playback instead.
```
lancesnider, line 480: Nit - in other parts of the doc, you do `#### scrub() (deprecated)`

lancesnider (approved):
One tiny nit, but feel free to merge if you disagree.

runtimes/web/rive-parameters.mdx
zplata: Yeah I had that initially but then realized its already under a H3 `Deprecated` section so I think its ok its omitted here

PR #914, opened 2026-09-01
docs(editor): globals binds are relative

lancesnider:
Documenting something that @pedroalpera resported here: https://2dimensions.slack.com/archives/C07MPAN5FDE/p1785742997402369

Globals use relative binding, which can result in naming conflicts.


bodymovin (approved):
LGTM

PR #917, opened 2026-09-02
rive(editor & support): reorg nav with best practices in mind

lancesnider:
- Remove `Workflows` from the top section
  - `Selection & Nav` and `Keyboard shortcuts` both move to `Editor Interface`
  - `Best Pactices` moves to its own section (see below)
  - `Dep graph`, `Tagging`, `Revision History` go into `Andvanced Topics/File Management`. These really didn't need to be before fundamentals. 
- `Fundamentals` doesn't change. 
- `Best Practices` - new top-level section. 
  - Makes `Accessibility` and other best practices more prominent. 
  - Add `Naming` page. 
- Advanced Topics
  - Move `Admin & Workspaces` to the support page. It makes more sense to have them next to other admin topics. 
  - `Legacy Features` - Sidebar is for nav we want people to go to. These legacy features were deprecated long enough ago that no reason to specifically drive people there. Search is enough. 

Before/After: 

CleanShot 2026-09-02 at 12 38 14@2x

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -86,8 +86,7 @@
                   "editor/keyboard-shortcuts",
                   "editor/workflows/dependency-graph",
                   "editor/tagging",
-                  "editor/fundamentals/revision-history",
-                  "getting-started/best-practices"
+                  "editor/fundamentals/revision-history"
                 ]
               }
             ]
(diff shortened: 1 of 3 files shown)


lancesnider:
I'm taking another crach at this in another PR.

PR #919, opened 2026-09-03
docs(scripting): shaders are live in production

lancesnider:
It's go-live day! We'll merge this when the time comes. 

- Update example code to not use `drawCanvas`
- Add a marketplace demo
- Remove EA note


HayesGordon (approved):
LGTM!

PR #924, opened 2026-09-03
docs(js+react): Add GPU canvas opt in docs

zplata:
Adding docs for GPU canvas opt in for JS + React

### DIFF
--- a/runtimes/react/gpu-canvas.mdx
+++ b/runtimes/react/gpu-canvas.mdx
@@ -37,7 +37,9 @@ export default function Scene() {
 
 The default exported `<Rive />` component takes the same prop:
 
-```jsx highlight={4}
+```jsx highlight={6}
+import Rive from '@rive-app/react-webgl2';
+
 <Rive
   src="my_file.riv"
   stateMachine="State Machine 1"
--- a/runtimes/react/gpu-canvas.mdx
+++ b/runtimes/react/gpu-canvas.mdx
@@ -51,7 +51,9 @@ GPU Canvas is disabled by default, but that may change in a future major version
 
 `useRiveFile` takes the same flag. A `RiveFile`'s rendering mode is fixed at creation, so changing `enableGPUCanvas` re-constructs the file:
 
-```jsx highlight={4,14}
+ ```jsx highlight={6,16}
+ import { useRive, useRiveFile } from '@rive-app/react-webgl2';
+ 
 export default function Scene() {
   const { riveFile, status } = useRiveFile({
     src: 'my_file.riv',
--- a/runtimes/react/gpu-canvas.mdx
+++ b/runtimes/react/gpu-canvas.mdx
@@ -51,9 +51,9 @@ GPU Canvas is disabled by default, but that may change in a future major version
 
 `useRiveFile` takes the same flag. A `RiveFile`'s rendering mode is fixed at creation, so changing `enableGPUCanvas` re-constructs the file:
 
- ```jsx highlight={6,16}
- import { useRive, useRiveFile } from '@rive-app/react-webgl2';
- 
+```jsx highlight={6,16}
+import { useRive, useRiveFile } from '@rive-app/react-webgl2';
+
 export default function Scene() {
   const { riveFile, status } = useRiveFile({
     src: 'my_file.riv',
--- a/runtimes/react/gpu-canvas.mdx
+++ b/runtimes/react/gpu-canvas.mdx
@@ -1,0 +1,92 @@
+---
+title: "GPU Canvas"
+description: "Opt a Rive component into GPU Canvas to render 3D content in React."
+---
+
+GPU Canvas renders 3D and shader content in the React runtime. It is opt-in per Rive instance and available in [`@rive-app/react-webgl2`](https://www.npmjs.com/package/@rive-app/react-webgl2) `4.34.0+`.
+
+<Warning>
+  GPU Canvas is experimental. The API may change in a minor release without a major version bump.
+</Warning>
+
+## Requirements
+
+- **`@rive-app/react-webgl2` only.** The Canvas2D packages (`@rive-app/react-canvas`, `@rive-app/react-canvas-lite`) do not support GPU Canvas yet.
+- **`useOffscreenRenderer` must stay false.** GPU Canvas rendering requires a WebGL context per `<canvas>`, while the offscreen renderer pattern shares one context across every canvas on the page. 
+  - React defaults this option to `true`, but enabling GPU Canvas flips that default to `false` for you, so there is nothing to change unless you set it to `true` explicitly.
+
+## Enabling GPU Canvas
+
+Pass `enableGPUCanvas: true` in the parameters you give `useRive`:
+
+```jsx highlight={9}
+import { useRive } from '@rive-app/react-webgl2';
+
+export default function Scene() {
+  const { RiveComponent } = useRive({
+    src: 'my_file.riv',
+    stateMachine: 'State Machine 1',
+    autoplay: true,
+    autoBind: true,
+    enableGPUCanvas: true,
+  });
+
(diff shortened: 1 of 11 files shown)


runtimes/web/caching-a-rive-file.mdx
```diff
@@ -77,4 +77,41 @@ window.addEventListener(
 },
 false
 );
-```
\ No newline at end of file
+```
+
+## GPU Canvas
```
lancesnider: This feels like more info than is needed by the user. The tldr is that caching GPU canvas content isn't gonna work and you're gonna get a warning, right? That feels like it could be a couple of paragraphs. I think it should also be a `<Note>`, rather than a top-level section. 

If we do want to dig into why, I'd make that a separate page. I would only link to this page from the note, rather than putting it in the sidebar.

runtimes/web/gpu-canvas.mdx
```diff
@@ -0,0 +1,75 @@
+---
+title: "GPU Canvas"
+description: "Opt a Rive instance into GPU Canvas to render 3D content on the web."
```
lancesnider:
```suggestion
description: "Opt a Rive instance into GPU Canvas to render 3D and shader content on the web."
```

runtimes/web/gpu-canvas.mdx
```diff
@@ -0,0 +1,75 @@
...
+---
+
+GPU Canvas renders 3D and shader content in the web runtime. It is opt-in per Rive instance and available in [`@rive-app/webgl2`](https://www.npmjs.com/package/@rive-app/webgl2) `2.42.0+`.
```
lancesnider:
```suggestion
GPU Canvas enables Rive files to render 3D and shader content in the web runtime. It is opt-in per Rive instance and available in [`@rive-app/webgl2`](https://www.npmjs.com/package/@rive-app/webgl2) `2.42.0+`.
```

runtimes/web/gpu-canvas.mdx
```diff
@@ -0,0 +1,75 @@
...
+  autoBind: true,
+  enableGPUCanvas: true,
+});
+```
+
+GPU Canvas is disabled by default, but that may change in a future major version.
+
+## Enabling GPU Canvas on a Cached File
+
+A `RiveFile` takes the same flag. Its rendering mode is fixed at import and has no setter, so set it when you construct the file:
```
lancesnider:
```suggestion
A `RiveFile` takes the same flag. The flag is fixed when the file is imported and has no setter, so set it when you construct the file:
```

runtimes/web/gpu-canvas.mdx
```diff
@@ -0,0 +1,75 @@
...
+  enableGPUCanvas: true,
+});
+```
+
+The file's flag wins over the same property on `new Rive()`: a `RiveFile` created with `enableGPUCanvas: true` renders through the required deferred renderer technique even when the `Rive` instance is created with `enableGPUCanvas: false` (or unset), and the runtime logs a warning.
+
+<Warning>
+  **A GPU Canvas `RiveFile` cannot be shared across Rive instances.** The renderer binds a file to a single `<canvas>`, so the first `Rive` instance you pass it to claims it.
+
+  Sharing one `RiveFile` among many instances does not fail the Rive load. The runtime will warn in the console, implicitly re-construct the `.riv` from the buffer the file already holds, and hand the additional `Rive` instance that fresh copy. Thus, the `RiveFile` you passed in goes unused, and nothing you read or configure through it applies. See [Caching a Rive file](/runtimes/web/caching-a-rive-file#gpu-canvas).
```
lancesnider:
```suggestion
Additional Rive instances will still load, but they won't use the `RiveFile` you passed in. Instead, the runtime logs a warning and creates a fresh copy of the `.riv` from the file's existing buffer. As a result, anything you read or configure through the original `RiveFile` does not apply to those additional instances.
```

runtimes/react/gpu-canvas.mdx
```diff
@@ -0,0 +1,96 @@
+---
+title: "GPU Canvas"
+description: "Opt a Rive component into GPU Canvas to render 3D content in React."
+---
+
+GPU Canvas renders 3D and shader content in the React runtime. It is opt-in per Rive instance and available in [`@rive-app/react-webgl2`](https://www.npmjs.com/package/@rive-app/react-webgl2) `4.34.0+`.
```
lancesnider:
```suggestion
GPU Canvas enables Rive files to render 3D and shader content in the React runtime. It is opt-in per Rive instance and available in [`@rive-app/react-webgl2`](https://www.npmjs.com/package/@rive-app/react-webgl2) `4.34.0+`.
```

runtimes/react/gpu-canvas.mdx
```diff
@@ -0,0 +1,96 @@
...
+  stateMachine="State Machine 1"
+  enableGPUCanvas
+/>
+```
+
+GPU Canvas is disabled by default, but that may change in a future major version.
+
+## Enabling GPU Canvas on a Cached File
+
+`useRiveFile` takes the same flag. A `RiveFile`'s rendering mode is fixed at creation, so changing `enableGPUCanvas` re-constructs the file:
```
lancesnider, line 54: I think I like this wording better than the equivalent you had in the web gpu-canvas page.

runtimes/react/gpu-canvas.mdx
```diff
@@ -0,0 +1,96 @@
...
+  return <RiveComponent />;
+}
+```
+
+The `RiveFile`'s flag wins over the same property on `useRive`. A `RiveFile` created with `enableGPUCanvas: true` renders through the required deferred renderer technique even when `useRive` is called with `enableGPUCanvas: false` (or unset), and the runtime logs a warning. `useRive` reads the file's mode too, so a GPU Canvas file turns the offscreen renderer optimization off, even when the component never set `enableGPUCanvas: true`.
+
+<Warning>
+  **A GPU Canvas file cannot be shared across Rive instances.** The renderer binds a file to a single `<canvas>`, so the first component you pass it to claims it. Call `useRiveFile` separately for each component that needs the same `.riv`.
+
+  Sharing one file among many components does not fail the Rive load. The runtime will warn in the console, implicitly re-construct the `.riv` from the buffer the file already holds, and hand the additional component that fresh copy. Thus, the file you passed in goes unused, and nothing you read or configure through it applies. See [Caching a Rive file](/runtimes/react/caching-a-rive-file#gpu-canvas).
```
lancesnider:
```suggestion
  Additional components will still load, but they won't use the file you passed in. Instead, the runtime logs a warning and creates a fresh copy of the `.riv` from the file's existing buffer. As a result, anything you read or configure through the original file does not apply to those additional components.
```

runtimes/react/gpu-canvas.mdx
```diff
@@ -0,0 +1,96 @@
...
+</Warning>
+
+## Limits
+
+Each GPU Canvas instance takes its own WebGL context, and browsers cap how many contexts a page can hold, so only opt in for the graphics that need it. See [WebGL Context Limits](/runtimes/web/canvas-vs-webgl#webgl-context-limits).
```
lancesnider:
```suggestion
Each GPU Canvas instance uses its own WebGL context, and browsers limit how many contexts a page can hold. Only enable GPU Canvas for graphics that need it. See [WebGL Context Limits](/runtimes/web/canvas-vs-webgl#webgl-context-limits).
```

Just matching the wording of the other doc.

runtimes/web/caching-a-rive-file.mdx
zplata: Made it a note: https://github.com/rive-app/rive-docs/pull/924/commits/97551c9cb99ca92bfd1322a21abeb9c7be898b63

lancesnider (approved):

PR #925, opened 2026-09-04
fix for wgsl basic shader example now in production

naychrist:
drawCanvas() was removed at some point. Script example can now be used as is again.


lancesnider:
I actually just merged another PR that fixes this, which is why I'm closing yours. Sorry for the confusion and thanks for contributing!

PR #929, opened 2026-09-04
fix: demo columns

lancesnider:
In production, this was making all the demos single column.


zplata:
@lancesnider not sure if this PR was ready for review, but just in case, the demos look single-column on the deploy link

toniarr (approved):

Issue #725, opened 2026-06-09
Editor docs contribution backlog


lancesnider:
This is a public list of documentation tasks that contributors can pick from.

## How to use this list

Comment on this issue before starting a task so we can help scope it and avoid duplicate work.

## General To Dos ✅ 

- [ ] When giving multi-step instructions, the gif should go _after_ the `<Steps>`. 

## Undocumented Features

- [x] Random state machine transitions - #775
- [x] React imperative method - #786 
- [x] Isolating objects  - #788
- [ ] 3d
- [x] Captions on state machines - #779
- [x] Default interpolation - #787
- [x] Scripting optimization and debug settings  (inspector when nothing selected)-  #793
- [x] Shaders, including shader targets (inspector when nothing selected) - #793
- [x] Interface viewport, including managing windows - #794
- [x] Sidebar reorder/columns - #788
- [x] Dependency Graph - #790
- [x] Projects - people confuse projects and workspaces - #802
- [x] Debug panel - It's no longer just a scripts thing - #796
- [x] Mesh weighting update - #811 
- [x] Drag to add corner radius or add more points - #834 
- [ ] Convert objects to ccomponent #913 
- [ ] Editing multiple selected keys at once

## Key

**Status:**
🟢 Done
🔵 Good enough
🟡 Needs work
🟠  Needs work - medium priority
🔴 Needs work - high priority

**Priority:**
💚  Good first issue
💛  Medium difficulty
❤️  Chat with staff before digging into these

Page | Notes | Status | Cost | Needs vid | Needs demo
-- | -- | -- | -- | -- | --
[getting-started/introduction](https://rive.app/docs/getting-started/introduction) | Use videos or .rivs instead of gifs. I'm not sure that's the perfect representation of a state machine. Maybe use walk, run, jump instead.  | 🔵 |  |
[editor/get-rive](https://rive.app/docs/editor/get-rive) |   | 🟢  |  |
[editor/selection-and-navigation](https://rive.app/docs/editor/interface-overview/selection-and-navigation) |   | 🟢  |  | | x
[editor/workflows/dependency-graph](https://rive.app/docs/editor/workflows/dependency-graph) |   | 🟢  |  | x | x
Interface Overview | Does this need re-doing now that we're redone all the child pages? Screenshots? | 🟠 |  |
[...]

PR #932, opened 2026-09-07
docs: improve text runs documentation

garimasrivastava26:
## Summary

- Clarify the parent-child relationship between Text objects and Text Runs.
- Add a marketplace example showing how Text Runs can be styled independently.

Related to #725


lancesnider:
I like that first paragraph. That makes runs much more clear. I updated the example a little to make it simpler.

lancesnider (approved):

PR #984, opened 2026-09-18
docs(apple): expand data binding and global asset coverage

dskuza:
## Summary

Expand the Apple concurrency API documentation to cover Font and Blob properties, nested View Model access and replacement, and the operations available for each property type. Explain shared asset lifetimes and correct the existing property examples.

Add a global asset capability table and native-font decoding examples. Include Blob asset metadata in the loading examples while keeping global registration limited to images, fonts, and audio.

## Validation

- Mintlify `broken-links`: no broken links
- MDX compilation passed for both updated pages
- `git diff --check` passed


HayesGordon (approved):
LGTM!

PR #986, opened 2026-09-18
docs(editor): eraser, brush, pencil drawing tools

lancesnider:


JcToon (changes requested):
A few notes:
- You can flatten and individual Spine. The Spine has  the option in the inspector, but when you click, all Spines of the Brush flatten. 

- We have two types of Simplify. One is a Erase option that it's on the document, and the other is i the edit vertex mode, where you can reduce the number of vertices. We can add this one in the flatten part of the document. 

- I miss the Shape Bend and the Brush Spine windows to edit the spines

lancesnider:
> * You can flatten and individual Spine. The Spine has  the option in the inspector, but when you click, all Spines of the Brush flatten.

Good call. Fixed. 

> * We have two types of Simplify. One is a Erase option that it's on the document, and the other is i the edit vertex mode, where you can reduce the number of vertices. We can add this one in the flatten part of the document.

👍 The second type of simplify isn't specific to the drawing tools so I added it to the Edit Vertices page: https://rive-freehand-drawing.mintlify.site/editor/fundamentals/edit-vertices#simplify

> * I miss the Shape Bend and the Brush Spine windows to edit the spines

The Shape Bend isn't specific to drawing tools - you can bend any path. I'm going to document this in another PR.

JcToon (approved):

